PR #2, opened 2022-06-18
Improve ssh analytics

jeanp413:
How to test for failure:
- SSH gateway:
  - To force the connection to fail we can remove all our ssh keys so you'll be required to use password authentication and then just type an incorrect password
- Local app:
  - ....

### DIFF
--- a/src/remoteConnector.ts
+++ b/src/remoteConnector.ts
@@ -726,7 +726,7 @@ export default class RemoteConnector extends Disposable {
 
 		await this.updateRemoteSSHConfig(usingSSHGateway, localAppSSHConfigPath);
 
-		await this.context.globalState.update(`ssh-remote+${sshDestination}`, { ...params, usingSSHGateway });
+		await this.context.globalState.update(sshDestination!, { ...params, usingSSHGateway });
 
 		vscode.commands.executeCommand(
 			'vscode.openFolder',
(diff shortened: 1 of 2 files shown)


akosyakov:
@jeanp413 It would be good to specify test cases 🙏 there are quite some cases, i.e. how do I check local app vs gateway, and then for both failure vs success?

jeanp413:
Added steps to test using gateway, not sure how to make a test for local app maybe just throw an error randomly :thinking:

akosyakov:
Do you test it again IDE lab project in Mixpanel?

andreafalzetti:
What do I need to do for testing this? Start a new workspace? 🤔

akosyakov:
> That the event is correctly send there? yeah or do you mean create a funnel?

@jeanp413 No, more like share with @andreafalzetti how he can verify it. A link in the PR description to some notion page would help.

akosyakov:
> check that the analytics event is fired successfully and only when the ssh connection is done by the remote-ssh extension. You need to clone this repo and hit f5 like any js project and then open a workspace and open in vscode desktop windows that is running the extension, if you have it installed delete it first so the dev version is used

We should document it somewhere and link from PR descriptions, like in CONTRIBUTING.md

jeanp413:
[Done](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/master/docs/CONTRIBUTING.md)

src/remoteConnector.ts
```diff
@@ -766,4 +760,53 @@ export default class RemoteConnector extends Disposable {
...
+			const [, sshDestStr] = remoteUri.authority.split('+');
+			const sshDest: { user: string; hostName: string } = JSON.parse(Buffer.from(sshDestStr, 'hex').toString('utf8'));
+			const gitpodHost = vscode.workspace.getConfiguration('gitpod').get<string>('host')!;
+			const host = new URL(gitpodHost).host;
+			const connectionSuccessful = sshDest.hostName.endsWith(host) && await isRemoteExtensionHostRunning();
+
+			const connectionInfo = this.context.globalState.get<SSHConnectionParams & { usingSSHGateway: boolean }>(sshDestStr);
+			if (connectionInfo) {
+				const kind = connectionInfo.usingSSHGateway ? 'gateway' : 'local-app';
+				if (connectionSuccessful) {
```
andreafalzetti, line 786: do we consider successful a connection even if the user failed to enter the right password?

jeanp413: Nope, it should send the `failed` status

andreafalzetti (approved):
tested for ssh gateway and works as advertised

PR #4, opened 2022-06-29
Add Heartbeat

jeanp413:
Move heartbeat to gitpod-desktop from gitpod-remote

How to test:
- [Run this extension](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/master/docs/CONTRIBUTING.md)
- Connect to a workspace in this prev env (use SSH gateway for now) https://jp-heartbeat-test.preview.gitpod-dev.com/workspaces (https://github.com/gitpod-io/gitpod/pull/11021)
    - make sure you have the debug VSCode Desktop window focus
- Install this gitpod-remote vsix [gitpod-remote-ssh-0.0.42.zip](https://github.com/gitpod-io/gitpod-vscode-desktop/files/9089328/gitpod-remote-ssh-0.0.42.zip) (related PR https://github.com/gitpod-io/openvscode-server/pull/378)
    - Upload `.zip` file to workspace, unzip it
    - Right-click on `.vsix` file and install it
- Check `__gitpod.cancelGitpodRemoteHeartbeat command executed` is logged in the gitpod-remote logs
- Open Gitpod with this PR https://github.com/gitpod-io/openvscode-server/pull/378
- Exec commands below to see if heartbeat sent
```
cd dev/gpctl
go build
./gpctl workspaces last-heartbeat <instance_id>
```
- Check only one close heartbeat is send

~~Not sure how to trigger ssh reconnection~~ :thinking: 

Maybe use `sudo killall sshd`, but window closed after sshd terminate, not sure if it works


jeanp413:
cc @akosyakov

src/extension.ts
```diff
@@ -95,7 +95,6 @@ export async function activate(context: vscode.ExtensionContext) {
 }
 
 export async function deactivate() {
-	if (telemetry) {
-		await telemetry.dispose();
-	}
+	await telemetry?.dispose();
```
akosyakov, line 98: I always wonder in such code if first dispose throws then second is never executed.

src/heartbeat.ts
```diff
@@ -0,0 +1,95 @@
+import * as vscode from 'vscode';
```
akosyakov, line 1: copyrights, maybe we can add some linting

src/remoteConnector.ts
```diff
@@ -622,6 +619,14 @@ export default class RemoteConnector extends Disposable {
 		throw new Error('SSH password modal dialog, Canceled');
 	}
 
+	private getGitpodSession() {
```
akosyakov, line 622: Should not it accept gitpodHost and always do a check 
```
if (new URL(params.gitpodHost).host !== new URL(gitpodHost).host) {
    ...
```
to be sure that we don't use session for wrong host

src/heartbeat.ts
```diff
@@ -0,0 +1,95 @@
...
+import { Disposable } from './common/dispose';
+import Log from './common/logger';
+import { withServerApi } from './internalApi';
+
+export class HeartbeatManager extends Disposable {
+
+    private lastActivity = new Date().getTime();
+    private gitpodHost = vscode.workspace.getConfiguration('gitpod').get<string>('host')!;
```
akosyakov, line 9: Should not we pass it via contructor to be sure? I would prefer that avoid using this setting, but provide value if we have it from params or other sources.
User can change this value for some reasons and then it will break us in weird ways. It would be enough to content the logic of checking and updating a setting in one place like `getGitpodSession` function.

src/remoteConnector.ts
```diff
@@ -810,12 +826,21 @@ export default class RemoteConnector extends Disposable {
 						gitpodHost: connectionInfo.gitpodHost
 					});
 				}
-				await this.context.globalState.update(remoteUri.authority, undefined);
+
+				// Don't await this on purpose so it doesn't block extension activation
+				this.startHeartBeat(connectionInfo);
```
akosyakov, line 831: We should do it as soon as possible avoid any unnecessary async requests and so on. It should not be matter whether we connected or remote extensio is running. Only that a window is opened. We should try to emualte logic from here:
https://github.com/gitpod-io/gitpod/blob/663e22a5501669080979387c96d0196e91df2834/components/supervisor/frontend/src/index.ts#L230-L241

src/heartbeat.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    private updateLastActivitiy() {
+        this.lastActivity = new Date().getTime();
+    }
+
+    private async sendHeartBeat(wasClosed?: true) {
+        const suffix = wasClosed ? 'was closed heartbeat' : 'heartbeat';
+        // if (wasClosed) {
+        this.logger.trace('sending ' + suffix);
```
akosyakov, line 81: we need to send `ide_close_signal` here https://www.notion.so/gitpod/ide_close_signal-e2a0f435fd4745b68567eb4d9afc8fd5

`clientKind` is `vscode`

akosyakov: It also implies that telemetry should be disposed after

src/remoteConnector.ts
```diff
@@ -761,6 +768,15 @@ export default class RemoteConnector extends Disposable {
 		}
 	}
 
+	private async startHeartBeat(connectionInfo: SSHConnectionParams) {
+		if (this.heartbeatManager?.instanceId !== connectionInfo.instanceId) {
```
akosyakov, line 772: I wonder why do we need thid guard? It should be called only once per window?

src/remoteConnector.ts
```diff
@@ -789,10 +805,10 @@ export default class RemoteConnector extends Disposable {
 			const usingSSHGateway = typeof sshDest !== 'string';
 
 			const connectionSuccessful = await isRemoteExtensionHostRunning();
-			const connectionInfo = this.context.globalState.get<SSHConnectionParams>(sshDestStr);
+			const connectionInfo = this.context.globalState.get<SSHConnectionParams & { isFirstConnection: boolean }>(sshDestStr);
 			if (connectionInfo) {
 				const kind = usingSSHGateway ? 'gateway' : 'local-app';
-				if (connectionSuccessful) {
+				if (connectionInfo.isFirstConnection && connectionSuccessful) {
 					this.telemetry.sendTelemetryEvent('vscode_desktop_ssh', {
```
akosyakov, line 812: I thought we reverthed this code, since it cannot be reliable? and put in previous place?

akosyakov:
> Not sure how to trigger ssh reconnection 🤔

I think you don't need to care here. The test case will be:
- observing that heartbeating is happening while window is opened and workspace is running
- if workspace is not running heartbeating should be stopped
- if window is closed close heartbeat should be send immediately, it should be preset in analytics as well

akosyakov:
I opened https://github.com/gitpod-io/gitpod/pull/11021 since I needed to rerun werft job.

src/remoteConnector.ts
```diff
@@ -726,12 +733,12 @@ export default class RemoteConnector extends Disposable {
...
 		await this.updateRemoteSSHConfig(usingSSHGateway, localAppSSHConfigPath);
 
-		await this.context.globalState.update(sshDestination!, params);
+		await this.context.globalState.update(sshDestination!, { ...params, isFirstConnection: true });
 
 		vscode.commands.executeCommand(
 			'vscode.openFolder',
 			vscode.Uri.parse(`vscode-remote://ssh-remote+${sshDestination}${uri.path || '/'}`),
-			{ forceNewWindow: true }
+			{ forceNewWindow: false } // REVERT THIS
```
akosyakov, line 741: `this.context.extensionMode == vscode.ExtensionMode.Production` ?

src/remoteConnector.ts
jeanp413: yes but just in case :slightly_smiling_face:

src/remoteConnector.ts
jeanp413: This should be async because it has to call getSession and for that the extension needs to be initialized. This happens because we have both gitpod auth and also require a session in the same extension.

src/extension.ts
jeanp413: mmm maybe we ensure dispose methods never throws

src/heartbeat.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    private updateLastActivitiy() {
+        this.lastActivity = new Date().getTime();
+    }
+
+    private async sendHeartBeat(wasClosed?: true) {
+        const suffix = wasClosed ? 'closed heartbeat' : 'heartbeat';
+        try {
+            await withServerApi(this.accessToken, this.gitpodHost, service => service.server.sendHeartBeat({ instanceId: this.instanceId, wasClosed }), this.logger);
+            this.telemetry.sendTelemetryEvent('ide_close_signal', { workspaceId: this.workspaceId, instanceId: this.instanceId, clientKind: 'vscode' });
```
akosyakov, line 94: only if closed

src/remoteConnector.ts
```diff
@@ -771,62 +791,92 @@ export default class RemoteConnector extends Disposable {
...
-			let decoded;
-			try {
-				decoded = JSON.parse(Buffer.from(sshDestStr, 'hex').toString('utf8'));
-			} catch {
-				// no-op
-			}
-			return decoded && typeof decoded.hostName === 'string' ? decoded : sshDestStr;
-		};
+		this.heartbeatManager = new HeartbeatManager(connectionInfo.gitpodHost, connectionInfo.workspaceId, connectionInfo.instanceId, accessToken, this.logger, this.telemetry);
+		this.logger.trace(`Heartbeat manager for workspace ${connectionInfo.workspaceId} (${connectionInfo.instanceId}) started`);
```
akosyakov, line 800: logging host makes sense that if someone shares with us we know whether it is related to gitpod.io

akosyakov:
It looks great, only tracking of workspace running state is missing that we don't send events if workspace is not running.

src/heartbeat.ts
```diff
@@ -0,0 +1,130 @@
...
+                const workspaceInfo = await service.server.getWorkspace(this.workspaceId);
+                this.isWorkspaceRunning = workspaceInfo.latestInstance?.status?.phase === 'running';
+                if (this.isWorkspaceRunning) {
+                    await service.server.sendHeartBeat({ instanceId: this.instanceId, wasClosed });
+                } else {
+                    this.stopHeartbeat();
+                }
+            }, this.logger);
+            if (wasClosed) {
+                this.telemetry.sendTelemetryEvent('ide_close_signal', { workspaceId: this.workspaceId, instanceId: this.instanceId, clientKind: 'vscode' });
```
akosyakov, line 108: we shoudl alway provide gitpodHost for telemetry from local clients, we shoudl be able to filter self hosted in analytics

src/heartbeat.ts
```diff
@@ -0,0 +1,130 @@
...
+            await withServerApi(this.accessToken, this.gitpodHost, async service => {
+                const workspaceInfo = await service.server.getWorkspace(this.workspaceId);
+                this.isWorkspaceRunning = workspaceInfo.latestInstance?.status?.phase === 'running';
+                if (this.isWorkspaceRunning) {
+                    await service.server.sendHeartBeat({ instanceId: this.instanceId, wasClosed });
+                } else {
+                    this.stopHeartbeat();
+                }
+            }, this.logger);
+            if (wasClosed) {
```
akosyakov, line 107: it should be executed only after `sendHeartBeat`. Since ide_close_signal is not sent otherwise.

src/heartbeat.ts
```diff
@@ -0,0 +1,130 @@
...
+    private updateLastActivitiy() {
+        this.lastActivity = new Date().getTime();
+    }
+
+    private async sendHeartBeat(wasClosed?: true) {
+        const suffix = wasClosed ? 'closed heartbeat' : 'heartbeat';
+        try {
+            await withServerApi(this.accessToken, this.gitpodHost, async service => {
+                const workspaceInfo = await service.server.getWorkspace(this.workspaceId);
+                this.isWorkspaceRunning = workspaceInfo.latestInstance?.status?.phase === 'running';
```
akosyakov, line 100: you should also check that instance id does not changed, as we discussed yesterday if a workspace is restarted then a user should go though browser again, we should not start sending heartbeat for restarted workspace

src/heartbeat.ts
```diff
@@ -0,0 +1,130 @@
...
+                this.updateLastActivitiy();
+                return null;
+            }
+        }));
+
+        this.logger.trace(`Heartbeat manager for workspace ${workspaceId} (${instanceId}) - ${gitpodHost} started`);
+
+        // Start heatbeating interval
+        this.sendHeartBeat();
+        this.heartBeatHandle = setInterval(() => {
```
akosyakov, line 81: interesting that this logic was changed in supervisor frontend to be less aggressive: https://github.com/gitpod-io/gitpod/blob/f157a029aeb9bcd37a29082dd13dc86a8c2d09e2/components/supervisor/frontend/src/ide/heart-beat.ts#L62-L71

I think we should align to reduce load as well. See https://github.com/gitpod-io/gitpod/pull/5613

src/remoteConnector.ts
```diff
@@ -771,62 +791,91 @@ export default class RemoteConnector extends Disposable {
...
-			} catch {
-				// no-op
-			}
-			return decoded && typeof decoded.hostName === 'string' ? decoded : sshDestStr;
-		};
+		this.heartbeatManager = new HeartbeatManager(connectionInfo.gitpodHost, connectionInfo.workspaceId, connectionInfo.instanceId, accessToken, this.logger, this.telemetry);
+	}
 
+	private async onGitpodRemoteConnection() {
 		const remoteUri = vscode.workspace.workspaceFile || vscode.workspace.workspaceFolders?.[0].uri;
```
akosyakov, line 803: I'm not sure about this check as it seems it can return undefine if workspace is slow and cannot load initial folders? Is it necessary?

src/remoteConnector.ts
```diff
@@ -771,62 +791,91 @@ export default class RemoteConnector extends Disposable {
...
+		const [, sshDestStr] = remoteUri.authority.split('+');
+		const connectionInfo = context.globalState.get<SSHConnectionParams & { isFirstConnection: boolean }>(`${RemoteConnector.SSH_DEST_KEY}${sshDestStr}`);
+
+		return !!connectionInfo;
+	}
+
+	return false;
+}
+
+// async function isRemoteExtensionHostRunning() {
```
akosyakov, line 863: they are already in git history, it should be safe to remove and you can always use blame to look them up?

akosyakov:
It looks almost there, some tweak are needed for heatbeating code. @mustard-mh @andreafalzetti since I'm away till next Monday, could you help to finish with testing after another round of changes? 🙏

src/remoteConnector.ts
jeanp413: Nope, it's still a proposed API, but `workspaceFile` or `workspaceFolders` is always defined even if ssh connection failed like using invalid password

src/remoteConnector.ts
```diff
@@ -771,62 +792,63 @@ export default class RemoteConnector extends Disposable {
...
+		if (!connectionInfo) {
+			return;
+		}
 
-			return connectionSuccessful;
+		await this.context.globalState.update(`${RemoteConnector.SSH_DEST_KEY}${sshDestStr}`, { ...connectionInfo, isFirstConnection: false });
+
+		// Check for gitpod remote extension version to avoid sending heartbeat in both extensions at the same time
+		const gitpodRemoteVersion = await getGitpodRemoteVersion();
+		if (gitpodRemoteVersion && semver.gte(gitpodRemoteVersion, '0.0.40')) {
```
jeanp413, line 819: `0.0.40` is a random value, will change it before merge

akosyakov:
I think logic looks well. Could someone help with testing and giving another look at it? 🙏 cc @mustard-mh @andreafalzetti

mustard-mh:
Tested, and it works well.

But extension requests auth twice when I'm testing. I'll update `How to Test` section later

```
gitpod /workspace/gitpod/dev/gpctl (jp/heartbeat-test) $ ./gpctl workspaces last-heartbeat 7fe830ee-af57-49bb-8c6d-91ad81cfc705
2022-07-11T08:54:24.876441322Z
gitpod /workspace/gitpod/dev/gpctl (jp/heartbeat-test) $ ./gpctl workspaces last-heartbeat 7fe830ee-af57-49bb-8c6d-91ad81cfc705
2022-07-11T08:56:18.843874055Z
gitpod /workspace/gitpod/dev/gpctl (jp/heartbeat-test) $ ./gpctl workspaces last-heartbeat 7fe830ee-af57-49bb-8c6d-91ad81cfc705
2022-07-11T08:56:18.843874055Z
gitpod /workspace/gitpod/dev/gpctl (jp/heartbeat-test) $ ./gpctl workspaces last-heartbeat 7fe830ee-af57-49bb-8c6d-91ad81cfc705
2022-07-11T08:56:48.681459343Z
```

mustard-mh:
After exec 

```
sudo killall sshd
```

The test VSCode Desktop window closed, don't know if `reconnect` works or not, since next time I open it, server maybe another one...

src/remoteConnector.ts
```diff
@@ -771,62 +792,63 @@ export default class RemoteConnector extends Disposable {
...
+			const session = await this.getGitpodSession(connectionInfo.gitpodHost);
```
mustard-mh, line 820: Should we use same session [above](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/2165d281beff7b17753b327266071b134479660a/src/remoteConnector.ts#L672)?

mustard-mh (approved):
code LGTM

src/remoteConnector.ts
jeanp413: Nope, vscode takes care of returning the same session if the scopes are the same, also this code will execute in a new window/after reload when it's connecting to a remote workspace, and the previous code only gets executed when opening a callback uri

akosyakov (approved):
🚀

jeanp413:
Could you do another test :pray: I updated the test steps
Changed it so gitpod-remote exposes a command to cancel the heartbeat and if this command returns successful then we start heartbeat on gitpod-desktop, this is safer as it could be possible to end up with many version combinations of both extensions as users could not always update both extensions at the same time
Also added a new telemetry event `vscode_desktop_heartbeat_state` so we can track if all new connections always activate the heartbeat in gitpod-desktop, If we have 100% rate and all events come from latest extension versions then we can remove the heartbeat code in gitpod-remote and also this telemetry event

mustard-mh:
We can disable computer's network to test `reconnect`. 😄

akosyakov:
> Also added a new telemetry event vscode_desktop_heartbeat_state so we can track if all new connections always activate the heartbeat in gitpod-desktop, If we have 100% rate and all events come from latest extension versions then we can remove the heartbeat code in gitpod-remote and also this telemetry event

sounds good, please update the tracking plan as usual 🙏

src/remoteConnector.ts
```diff
@@ -771,62 +791,64 @@ export default class RemoteConnector extends Disposable {
...
+		const connectionInfo = this.context.globalState.get<SSHConnectionParams & { isFirstConnection: boolean }>(`${RemoteConnector.SSH_DEST_KEY}${sshDestStr}`);
+		if (!connectionInfo) {
+			return;
+		}
 
-			return connectionSuccessful;
+		await this.context.globalState.update(`${RemoteConnector.SSH_DEST_KEY}${sshDestStr}`, { ...connectionInfo, isFirstConnection: false });
+
+		// Check for gitpod remote extension version to avoid sending heartbeat in both extensions at the same time
+		const isGitpodRemoteHeartbeatCancelled = await cancelGitpodRemoteHeartbeat();
```
akosyakov, line 817: I'm not sure about it. We see that remote-ssh-extension fails to install right now very often. If we do it then heartbeat is missing.

andreafalzetti:
I am having issues to connect to the workspace. Any idea? 🤔 

Screenshot 2022-07-12 at 13 38 07

jeanp413:
You need to create a new prev env, the one I posted got merged

andreafalzetti (approved):

PR #5, opened 2022-07-09
Use new `getSSHPublicKeys` api method

jeanp413:
## Description

Improve ssh connection by using the new `getSSHPublicKeys` api method, owner token is no longer needed as part of the remote uri authority.

## How to test

1. [Run this extension](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/master/docs/CONTRIBUTING.md)
2. Create a workspace in this prev env https://jp-vscode-ab4260534a.preview.gitpod-dev.com/workspaces (https://github.com/gitpod-io/gitpod/pull/11246) and connect to vscode desktop
3. If you don't have any SSH keys registered in your gitpod account then it will tell you to use the ownertoken
4. Register an SSH key in your gitpod account, and try connecting again, it should connect directly

Questions:
- Should we keep previous behavior and add a migration path to the new behavior? As it is right now all current users connecting SSH gateway (not that many) will be forced to connect using ownertoken until they register an SSH key in their Gitpod account.

## Release Notes

```release-note
NONE
```


src/remoteConnector.ts
```diff
@@ -615,7 +632,7 @@ export default class RemoteConnector extends Disposable {
 
 		const copy = 'Copy';
 		const configureSSH = 'Configure SSH';
-		const action = await vscode.window.showInformationMessage(`An SSH key is required for passwordless authentication.\nAlternatively, copy and use this password: ${maskedPassword}`, { modal: true }, copy, configureSSH);
+		const action = await vscode.window.showInformationMessage(`Couldn't find any SSH public key registered in your Gitpod account.\nAlternatively, copy and use this password: ${maskedPassword}`, { modal: true }, copy, configureSSH);
```
akosyakov, line 635: `You don't have any public SSH keys in your Gitpod account. You can [add a new public key](https://gitpod.io/keys), or use a generated access token as a password.`

jeanp413: We can't use markdown in modal notification

jeanp413: Change it to `You don't have registered any SSH public key for this machine in your Gitpod account.\nAlternatively, copy and use this password:`

jeanp413:
Awaiting until https://github.com/gitpod-io/gitpod/pull/11246 is deployed

akosyakov:
> Should we keep previous behavior and add a migration path to the new behavior? As it is right now all current users connecting SSH gateway (not that many) will be forced to connect using ownertoken until they register an SSH key in their Gitpod account.

I think we don't need to care much. We don't have many users for SSH Gateway yet and they are using experimental feature as for now.

src/remoteConnector.ts
```diff
@@ -615,13 +632,13 @@ export default class RemoteConnector extends Disposable {
 
 		const copy = 'Copy';
 		const configureSSH = 'Configure SSH';
-		const action = await vscode.window.showInformationMessage(`An SSH key is required for passwordless authentication.\nAlternatively, copy and use this password: ${maskedPassword}`, { modal: true }, copy, configureSSH);
+		const action = await vscode.window.showInformationMessage(`You don't have registered any SSH public key for this machine in your Gitpod account.\nAlternatively, copy and use this password: ${maskedPassword}`, { modal: true }, copy, configureSSH);
```
akosyakov, line 635: Maybe let's change to `showWarningMessage`.

Users should feel that it is not OK to use password and they should move to use SSH keys.

I think we should be also specific that password is generated for each workspace instance. Can we add something like `Password resets on every workspace restart.`?

src/remoteConnector.ts
```diff
@@ -494,8 +498,28 @@ export default class RemoteConnector extends Disposable {
...
+				return false;
+			}
+
+			const parsedKey = Array.isArray(parsedResult) ? parsedResult[0] : parsedResult;
+			const fingerprint = crypto.createHash('sha256').update(parsedKey.getPublicSSH()).digest('base64');
+			return keyFingerprints.includes(fingerprint);
+		});
+		this.logger.trace(`Registered public keys in Gitpod account:`, identityFilePaths.length ? identityFilePaths.toString() : 'None');
 
 		// Commented this for now as `checkDefaultIdentityFiles` seems enough
```
akosyakov, line 524: (nit) it is in git history already?

akosyakov (approved):
Code wise it looks good.

@iQQBot Could you help to test it please? 🙏

iQQBot (approved):
Test on prod, looks good

PR #6, opened 2022-07-12
Fix auth validation

jeanp413:
## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11296

## How to test

1. [Run this extension](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/master/docs/CONTRIBUTING.md)
2. Connect to a workspace in any prev env (use SSH gateway) you can use this one https://jp-heartbeat-test.preview.gitpod-dev.com/workspaces
3. After successful connection, delete the gitpod token from database, or you can rebuild the prev env with `with-clean-slate-deployment` so DB gets recreated
4. Try to connect again to a workspace it should ask to login again and not fail silently as before

## Release Notes

```release-note
NONE
```

### DIFF
--- a/src/authentication.ts
+++ b/src/authentication.ts
@@ -144,6 +144,7 @@ export default class GitpodAuthenticationProvider extends Disposable implements 
 				if (e.message === 'Unexpected server response: 401') {
 					return undefined;
 				}
+				this._logger.error(`Error while verifying session with the following scopes: ${scopesStr}`, e);
 			}
 
 			this._logger.trace(`Read the following session from the keychain with the following scopes: ${scopesStr}`);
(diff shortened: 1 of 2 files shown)


src/authentication.ts
```diff
@@ -136,15 +136,13 @@ export default class GitpodAuthenticationProvider extends Disposable implements
...
-				} catch (e) {
-					// Remove sessions that return unauthorized response
-					if (e.message === 'Unauthorized') {
-						return undefined;
-					}
+			try {
+				userInfo = await this._gitpodServer.getUserInfo(session.accessToken);
+				this._logger.info(`Verified session with the following scopes: ${scopesStr}`);
+			} catch (e) {
+				// Remove sessions that return unauthorized response
```
akosyakov, line 143: It seem we are swallowing errors here, i think we should log them at least

andreafalzetti:
is any prev env good for testing this? the one linked doesn't work anymore I think

jeanp413:
@andreafalzetti check slack

andreafalzetti (approved):
Tested and works as advertised

PR #7, opened 2022-07-14
Add in product changelog

mustard-mh:
## TODO
- [x] Use some endpoints (discussing) to query `latest publish date` and `markdown content`
- [x] Aligh to https://github.com/gitpod-io/openvscode-server/pull/381
- [x] Remove debug command

## Related Issue(s)

Fixes #

## How to test

- Follow steps in https://github.com/gitpod-io/gitpod-vscode-desktop/blob/master/docs/CONTRIBUTING.md
- Press `F5`
- Toggle command `Gitpod: Show Release Note`

## Release Notes

```release-note
NONE
```

## Incase you don't want to run it

|Screenshot| Light|
|-|-|
|image|image|

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
 		"onCommand:gitpod.syncProvider.remove",
 		"onCommand:gitpod.exportLogs",
 		"onCommand:gitpod.api.autoTunnel",
+		"onCommand:gitpod.showReleaseNotes",
 		"onAuthenticationRequest:gitpod",
 		"onUri"
 	],
@@ -92,6 +93,11 @@
 				"command": "gitpod.exportLogs",
 				"category": "Gitpod",
 				"title": "Export all logs"
+			},
+			{
+				"command": "gitpod.showReleaseNotes",
+				"category": "Gitpod",
+				"title": "Show Release Notes"
 			}
 		]
 	},
@@ -109,6 +115,7 @@
 		"@types/analytics-node": "^3.1.9",
 		"@types/crypto-js": "4.1.1",
 		"@types/google-protobuf": "^3.7.4",
+		"@types/js-yaml": "^4.0.5",
 		"@types/node": "16.x",
 		"@types/node-fetch": "^2.5.12",
 		"@types/semver": "^7.3.10",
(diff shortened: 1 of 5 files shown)


jeanp413:
Could you add some CSS magic :lipstick: 
image

mustard-mh:
> Could you add some CSS magic 💄

@jeanp413  It's a TODO 🙈, do you have any suggested markdown CSS?

jeanp413:
lets's try first using the same css as vscode release notes an then we can iterate from there if necessary, check `src/vs/workbench/contrib/markdown/browser/markdownDocumentRenderer.ts` in vscode

jeanp413:
Nice :ok_hand:, not sure about the image because it takes to much screen space but other than that it look really good

jeanp413:
@loujaybee @akosyakov thoughs about the image, should we keep it or not?

package.json
```diff
@@ -87,6 +89,16 @@
...
 				"title": "Export all logs"
+			},
+			{
+				"command": "gitpod.releaseNote",
+				"category": "Gitpod",
+				"title": "Show Release Note"
+			},
+			{
+				"command": "gitpod.cleanReleaseNoteCache",
+				"category": "Gitpod",
```
akosyakov, line 100: Debug actions should go under `Developer`

package.json
```diff
@@ -87,6 +89,16 @@
 				"command": "gitpod.exportLogs",
 				"category": "Gitpod",
 				"title": "Export all logs"
+			},
+			{
+				"command": "gitpod.releaseNote",
```
akosyakov, line 94: `showReleaseNotes`

akosyakov:
> @loujaybee @akosyakov thought about the image, should we keep it or not?

I think updates to the content should be done directly in Changelog. I would like to see list of highlights and call out to use latest versions in each changelog at the top. Images should go to specific highlights below. cc @Cahllagerfeld @mikenikles

package.json
mustard-mh: I will remove it before merge

akosyakov:
We should use https://code.visualstudio.com/api/advanced-topics/remote-extensions#sync-user-global-state-between-machines to ensure that a user don’t see already seen changelog across different machines (local and Gitpod workspaces) [Example](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#setkeysforsync-example)

PR #10, opened 2022-07-27
Add telemetry event for ssh modal

mustard-mh:
## Description

Same as title, to see how many users go with copy password (no SSH public key uploaded yet)

See also [internal chat](https://gitpod.slack.com/archives/C01KGM9BH54/p1658826643857369?thread_ts=1658825096.949629&cid=C01KGM9BH54)

## TODO
- [ ] Add [tracking plan](https://www.notion.so/gitpod/d1c05025dae64f4e8dcd3188f121e270?v=e7e87f95dcac436f86d4ec2698e795a7) in notion

## How to test

No need

### DIFF
--- a/src/remoteConnector.ts
+++ b/src/remoteConnector.ts
@@ -618,10 +618,10 @@ export default class RemoteConnector extends Disposable {
 		return true;
 	}
 
-	private async showSSHPasswordModal(password: string, gitpodHost: string, sshParams: SSHConnectionParams) {
+	private async showSSHPasswordModal(password: string, sshParams: SSHConnectionParams) {
 		const maskedPassword = '•'.repeat(password.length - 3) + password.substring(password.length - 3);
 
-		const gitpodVersion = await getGitpodVersion(gitpodHost);
+		const gitpodVersion = await getGitpodVersion(sshParams.gitpodHost);
 		const sshKeysSupported = isFeatureSupported(gitpodVersion, 'SSHPublicKeys');
 
 		const copy: vscode.MessageItem = { title: 'Copy' };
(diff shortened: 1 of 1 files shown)


src/remoteConnector.ts
```diff
@@ -708,6 +708,7 @@ export default class RemoteConnector extends Disposable {
 
 				if (password) {
 					await this.showSSHPasswordModal(password, params.gitpodHost);
+					this.telemetry.sendRawTelemetryEvent('vscode_desktop_ssh', { kind: 'gateway', status: 'connecting', ...params, gitpodVersion: gitpodVersion.raw, copy: true });
```
akosyakov, line 711: I don't think we should reuse `vscode_desktop_ssh`,

but instead have a new event like `vscode_desktop_ssh_gateway_modal` with all metadata as for `vscode_desktop_ssh` event + `action` value plus `cancelled` for undefined.

src/remoteConnector.ts
```diff
@@ -631,6 +631,7 @@ export default class RemoteConnector extends Disposable {
 			? `You don't have registered any SSH public key for this machine in your Gitpod account.\nAlternatively, copy and use this temporary password until workspace restart: ${maskedPassword}`
 			: `An SSH key is required for passwordless authentication.\nAlternatively, copy and use this password: ${maskedPassword}`;
 		const action = await vscode.window.showWarningMessage(message, { modal: true }, copy, configureSSH, showLogs);
+        this.telemetry.sendRawTelemetryEvent('vscode_desktop_ssh_gateway_modal', { kind: 'gateway', action: action?.title ?? 'Cancelled', ...sshParams, gitpodVersion: gitpodVersion.raw });
```
jeanp413, line 634: `cancelled` and `configureSSH` options telemetry info is already send in line 716

src/remoteConnector.ts
```diff
@@ -707,7 +708,7 @@ export default class RemoteConnector extends Disposable {
 				sshDestination = destination;
 
 				if (password) {
-					await this.showSSHPasswordModal(password, params.gitpodHost);
+					await this.showSSHPasswordModal(password, params.gitpodHost, params);
 				}
 
 				this.telemetry.sendRawTelemetryEvent('vscode_desktop_ssh', { kind: 'gateway', status: 'connected', ...params, gitpodVersion: gitpodVersion.raw });
```
jeanp413, line 714: If we need to know if user selected `password` then just add another property here

mustard-mh: is this (add `password` property) good for you? @akosyakov

mustard-mh: We only know user copied the `password` but don't know if user paste and connect it. So maybe this tracking event's `status=connected` is not correct?

akosyakov: We are going to remove errors from vscode_desktop_ssh eventually and log it to GCP. We should not rely on it. We should have one event which shows us a usage of this notification, i.e. captures all exits.

src/remoteConnector.ts
akosyakov, line 634: `kind: 'gateway'` can be dropped, the modal only about connectivity with SSH Gateway

akosyakov (approved):
please address https://github.com/gitpod-io/gitpod-vscode-desktop/pull/10/files#r933172270

it looks good to me otherwise

We also should do following:
- update the tracking plan
- release it for latest and insiders

@jeanp413 can help with the release, since there are some subtleties with useLocalApp

PR #11, opened 2022-07-29
Get ssh identity file from ssh config and ssh agent

mustard-mh:
## Description

Support default ssh behaviors of IdentityFile

SSH Config code is a copy of https://github.com/jeanp413/open-remote-ssh/blob/master/src/ssh/sshConfig.ts

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11467

## How to test

- Follow this doc to run extension locally https://github.com/gitpod-io/gitpod-vscode-desktop/blob/master/docs/CONTRIBUTING.md
- Make sure debug vscode window is actived
- Create a workspace and open with desktop vscode, it should back to the debug window
- Start test with cases below

### IdentityFile
- Setup your `~/.ssh/config` (mac) with gitpod host and custom identity file
- Add an ssh key to the ssh agent with `ssh-add`
- Run extension with this branch
- Open Gitpod workspace with Desktop VSCode and connect using your debug vscode window
- See if your custom identify file works

### IdentityAgent
- Follow the [doc](https://developer.1password.com/docs/ssh/get-started#step-3-turn-on-the-1password-ssh-agent) to add and turn on 1Password ssh-agent, i.e. `id_ed25519`
```
# you can gen one by command below, which is recommend by 1password
ssh-keygen -t ed25519 -b 4096
```
- Remove related ssh key files in your computer. i.e. `id_ed25519` `id_ed25519.pub`
- Upload that ssh key to https://gitpod.io/keys and make sure only that one exists in https://gitpod.io/keys
- Run extension locally and make sure debug one is activated
- Open a new workspace with desktop VSCode
- Check if we can access without copy password modal

### Errors
Fill up incorrect `identityFile` and `identityAgent`

### DIFF
--- a/src/common/files.ts
+++ b/src/common/files.ts
@@ -5,7 +5,6 @@
 
 import * as fs from 'fs';
 import * as os from 'os';
-import * as path from 'path';
 
 const homeDir = os.homedir();
 
@@ -20,18 +19,4 @@ export async function exists(path: string) {
 
 export function untildify(path: string){
 	return path.replace(/^~(?=$|\/|\\)/, homeDir);
-}
-
-export function resolveHomeDir(filepath: string | undefined) {
-    if (!filepath) {
-        return filepath;
-    }
-    const homedir = os.homedir();
-    if (filepath === '~') {
-        return homedir;
-    }
-    if (filepath.startsWith('~/')) {
-        return path.join(homedir, filepath.slice(2));
-    }
-    return filepath;
 }
--- a/src/common/files.ts
+++ b/src/common/files.ts
@@ -5,6 +5,7 @@
 
 import * as fs from 'fs';
 import * as os from 'os';
+import * as path from 'path';
 
 const homeDir = os.homedir();
 
@@ -19,4 +20,18 @@ export async function exists(path: string) {
 
 export function untildify(path: string){
 	return path.replace(/^~(?=$|\/|\\)/, homeDir);
+}
+
+export function resolveHomeDir(filepath: string | undefined) {
+    if (!filepath) {
+        return filepath;
+    }
+    const homedir = os.homedir();
+    if (filepath === '~') {
+        return homedir;
+    }
+    if (filepath.startsWith('~/')) {
+        return path.join(homedir, filepath.slice(2));
+    }
+    return filepath;
 }
--- a/src/common/files.ts
+++ b/src/common/files.ts
@@ -4,6 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as fs from 'fs';
+import * as os from 'os';
+
+const homeDir = os.homedir();
 
 export async function exists(path: string) {
     try {
@@ -12,4 +15,8 @@ export async function exists(path: string) {
     } catch {
         return false;
     }
+}
+
+export function untildify(path: string){
+	return path.replace(/^~(?=$|\/|\\)/, homeDir);
 }
--- a/src/remoteConnector.ts
+++ b/src/remoteConnector.ts
@@ -464,7 +464,7 @@ export default class RemoteConnector extends Disposable {
 
 		let sshAgentParsedKeys: ParsedKey[] = [];
 		try {
-			let sshAgentSock = isWindows ? `\\.\pipe\openssh-ssh-agent` : (hostConfig['IdentityAgent'] || process.env['SSH_AUTH_SOCK']);
+			let sshAgentSock = isWindows ? '\\\\.\\pipe\\openssh-ssh-agent' : (hostConfig['IdentityAgent'] || process.env['SSH_AUTH_SOCK']);
 			if (!sshAgentSock) {
 				throw new Error(`SSH_AUTH_SOCK environment variable not defined`);
 			}
--- a/src/remoteConnector.ts
+++ b/src/remoteConnector.ts
@@ -28,7 +28,7 @@ import { HeartbeatManager } from './heartbeat';
 import { getGitpodVersion, isFeatureSupported } from './featureSupport';
 import SSHConfiguration from './ssh/sshConfig';
 import { isWindows } from './common/platform';
-import { resolveHomeDir, untildify } from './common/files';
+import { untildify } from './common/files';
 
 interface SSHConnectionParams {
 	workspaceId: string;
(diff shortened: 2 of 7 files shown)


src/remoteConnector.ts
```diff
@@ -504,7 +515,7 @@ export default class RemoteConnector extends Disposable {
 			this.logger.error(`Couldn't write '${sshDestInfo.hostName}' host to known_hosts file:`, e);
 		}
 
-		let identityFilePaths = await checkDefaultIdentityFiles();
+		let identityFilePaths = await this.getIdentityFilePaths(gitpodHost);
```
jeanp413, line 518: I think it would be better to use `sshDestInfo.hostName` if there's a rule `*.gitpod.io` the ssh-config library will match it too, could you try that

src/remoteConnector.ts
```diff
@@ -429,6 +430,16 @@ export default class RemoteConnector extends Disposable {
 		}
 	}
 
+	private async getIdentityFilePaths(gitpodHost: string) {
+		const list = await checkDefaultIdentityFiles();
+		const sshConfiguration = await SSHConfiguration.loadFromFS();
+		const hostConfiguration = sshConfiguration.getHostConfiguration(new URL(gitpodHost).host);
+		if (!!hostConfiguration.IdentityFile) {
+			list.unshift(hostConfiguration.IdentityFile);
```
jeanp413, line 438: from the docs `hostConfiguration.IdentityFile` is an array, can it also be a string? if so we should validate it and make it an array 
We also need to validate if the file exists

jeanp413:
@mustard-mh I also added logic to read ssh key from the ssh agent so we can completely close https://github.com/gitpod-io/gitpod/issues/11467

akosyakov:
Please add [[1](https://github.com/gitpod-io/gitpod/issues/11467#issuecomment-1196401154)] to `How to test` to prove that code is reliable for all kind of use cases. If not we need a way to enable not captured use cases.

akosyakov:
Also cc @loujaybee we will need to update docs and the announcement blog about expectations.

src/remoteConnector.ts
mustard-mh: Note: With [ssh config doc](https://www.ssh.com/academy/ssh/config), `IdentityFile` is only a file, but with [ssh-config](https://github.com/cyjake/ssh-config#compute-parameters-by-host) module, it'll return an array after `compute`

mustard-mh:
I updated **IdentityAgent** in **How to test** section. cc @akosyakov

mustard-mh:
Could you help test and review 🙏 ? @felladrin @iQQBot

src/remoteConnector.ts
jeanp413: yeah that's what I meant about [ssh-config](https://github.com/cyjake/ssh-config#compute-parameters-by-host) module, it's already fixed when I did my changes

src/common/files.ts
```diff
@@ -20,3 +21,17 @@ export async function exists(path: string) {
...
+    }
+    const homedir = os.homedir();
+    if (filepath === '~') {
+        return homedir;
+    }
+    if (filepath.startsWith('~/')) {
+        return path.join(homedir, filepath.slice(2));
+    }
+    return filepath;
+}
```
jeanp413, line 37: This is the what `untildify` does, does it not work?

src/remoteConnector.ts
```diff
@@ -440,7 +440,7 @@ export default class RemoteConnector extends Disposable {
 			identityFiles.push(...DEFAULT_IDENTITY_FILES);
 		}
 
-		const identityFileContentsResult = await Promise.allSettled(identityFiles.map(async path => fs.promises.readFile(path + '.pub')));
+		const identityFileContentsResult = await Promise.allSettled(identityFiles.map(async path => fs.promises.readFile(resolveHomeDir(path) + '.pub')));
```
jeanp413, line 443: already done in line 438 and  DEFAULT_IDENTITY_FILES are resolved already

src/remoteConnector.ts
```diff
@@ -462,14 +462,20 @@ export default class RemoteConnector extends Disposable {
 			};
 		}).filter(<T>(v: T | undefined): v is T => !!v);
 
-		const sshAgentSock = isWindows ? `\\.\pipe\openssh-ssh-agent` : (hostConfig['IdentityAgent'] || process.env['SSH_AUTH_SOCK']);
```
jeanp413: just use `untildify`

jeanp413:
@mustard-mh I think latest [commit](https://github.com/gitpod-io/gitpod-vscode-desktop/pull/11/commits/9e4f9a7c3953638d62c3a402bb4b3067924f374e) is not necessary so would say just revert it and do a 1 line change in line 645 adding the `untildify` call

jeanp413:
I'll go ahead and address the comments as I want to test on windows as I won't be available tomorrow/friday because holidays :slightly_smiling_face:

mustard-mh:
> @mustard-mh I think latest [commit](https://github.com/gitpod-io/gitpod-vscode-desktop/pull/11/commits/9e4f9a7c3953638d62c3a402bb4b3067924f374e) is not necessary so I would say just revert it and do a 1 line change in line 465 adding the untildify call

Ahh, just realized that `untildify` is working for this 🌴

mustard-mh:
I tested (MacOS) with `IdentityFile` and `IdentityAgent`, works well.

Note that:
- `.pub` file can be deleted, fingerprint calc in default ssh behavior works well, but not in our extension because we use `.pub` file to calc the fingerprint
- Tried not to use `.pub` but the package we are using (`ssh2`) does not support calc fingerprint for `private key with password`, maybe find another package that can resolve it

They are not the reasons to block this PR, but we can mark it

jeanp413:
I just finished testing in windows, did a fix and it's working 🎉 

> .pub file can be deleted, fingerprint calc in default ssh behavior works well, but not in our extension because we use .pub file to calc the fingerprint

what's the point of deleting the .pub file?

mustard-mh:
> what's the point of deleting the .pub file?

No, just said it can be deleted, and in case users have this habit (😅) and face it in the future

jeanp413:
> I tested (MacOS) with IdentityFile and IdentityAgent, works well.

I tested on windows and linux and looks there are no regression, so I'll go ahead and merge it and release and a pre-release version, I'll ask tarun to try it (he's using with 1password integration) and a user in discord who is reporting some issues with previous behavior

PR #13, opened 2022-08-05
Use min version once version api unreachable

mustard-mh:
## Related Issue(s)

Fixes #

## How to test

Change gitpod host to incorrect one, see if it failed with version `0.0.0`

### DIFF
--- a/src/featureSupport.ts
+++ b/src/featureSupport.ts
@@ -4,16 +4,21 @@
  *--------------------------------------------------------------------------------------------*/
 import * as semver from 'semver';
 import fetch from 'node-fetch';
+import Log from './common/logger';
 
 export class GitpodVersion {
-    static DEFAULT_VERSION = '9999.99.99';
+    static MAX_VERSION = '9999.99.99';
+    static MIN_VERSION = '0.0.0';
+
+    static Max = new GitpodVersion(GitpodVersion.MAX_VERSION);
+    static Min = new GitpodVersion(GitpodVersion.MIN_VERSION);
 
     readonly version: string;
     readonly raw: string;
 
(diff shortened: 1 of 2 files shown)


src/featureSupport.ts
```diff
@@ -32,39 +37,59 @@ export class GitpodVersion {
...
+            } else {
+                return undefined;
+            }
         }
+    };
 
-        gitpodRawVersion = await versionResponse.text();
-    } catch (e) {
-        return undefined;
+    const gitpodRawVersion = await fetchVersion();
```
jeanp413, line 72: if `gitpodRawVersion` is undefined then fallback to GitpodVersion.Min

src/featureSupport.ts
```diff
@@ -32,39 +37,59 @@ export class GitpodVersion {
...
+    if (!gitpodRawVersion) {
+        return {
```
jeanp413, line 74: Maybe add one more log here telling we are using min version as default

jeanp413 (approved):

src/featureSupport.ts
mustard-mh: Added

mustard-mh:
It's not correct!! @jeanp413 , the static string have no 'release-' prefix 😱

image

mustard-mh:
I'm going to push directly to fix it, no we publish a lot of versions

mustard-mh:
No, my test cases are wrong, we have code below, if add `release-` prefix, it's not correct because it will remove last number
```
        if (semver.valid(gitpodVersion)) {
            this.version = gitpodVersion;
        }
```

mustard-mh:
I'm going to create a unit test PR for it 🥲

PR #16, opened 2022-08-22
Initialize remote extensions

jeanp413:
## Description

Initialize remote extensions with extensions from sync server

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/7906
Related https://github.com/gitpod-io/openvscode-server/pull/419

## How to test

1. Run this extension and connect to workspace 
2. Install this vsix [gitpod-remote-ssh-0.0.39.zip](https://github.com/gitpod-io/gitpod-vscode-desktop/files/9392713/gitpod-remote-ssh-0.0.39.zip) so it's installed by default the second time
3. Disconnect from workspace
4. Enable settings sync
5. Connect to workspace again and check sync extensions are installed in remote server

### DIFF
--- a/src/remoteConnector.ts
+++ b/src/remoteConnector.ts
@@ -941,7 +941,13 @@ export default class RemoteConnector extends Disposable {
 					vscode.commands.executeCommand('workbench.userDataSync.actions.turnOn');
 				}
 			} else {
-				this.logger.error('Error while fetching settings sync extension data', e);
+				this.logger.error('Error while fetching settings sync extension data:', e);
+
+				const seeLogs = 'See Logs';
+				const action = await vscode.window.showErrorMessage(`Error while fetching settings sync extension data.`, seeLogs);
+				if (action === seeLogs) {
+					this.logger.show();
+				}
 			}
 			return;
 		}
@@ -966,12 +972,26 @@ export default class RemoteConnector extends Disposable {
 		}
 
 		try {
-			this.logger.trace(`Try installing extensions on remote: `, extensions.map(e => e.identifier.id).join('\n'));
-			await retry(async () => {
-				await vscode.commands.executeCommand('__gitpod.initializeRemoteExtensions', extensions);
-			}, 3000, 15);
+			await vscode.window.withProgress<void>({
+				title: 'Installing extensions on remote',
+				location: vscode.ProgressLocation.Notification
+			}, async () => {
+				try {
+					this.logger.trace(`Installing extensions on remote: `, extensions.map(e => e.identifier.id).join('\n'));
+					await retry(async () => {
+						await vscode.commands.executeCommand('__gitpod.initializeRemoteExtensions', extensions);
+					}, 3000, 15);
+				} catch (e) {
+					this.logger.error(`Could not execute '__gitpod.initializeRemoteExtensions' command`);
+					throw e;
+				}
+			});
 		} catch {
-			this.logger.error(`Could not execute '__gitpod.initializeRemoteExtensions' command`);
+			const seeLogs = 'See Logs';
+			const action = await vscode.window.showErrorMessage(`Error while installing extensions on remote.`, seeLogs);
+			if (action === seeLogs) {
+				this.logger.show();
+			}
 		}
 	}
 
--- a/src/remoteConnector.ts
+++ b/src/remoteConnector.ts
@@ -1031,7 +1031,10 @@ export default class RemoteConnector extends Disposable {
 			this.logger.warn(`Local heatbeat not supported in ${connectionInfo.gitpodHost}, using version ${gitpodVersion.version}`);
 		}
 
-		this.initializeRemoteExtensions();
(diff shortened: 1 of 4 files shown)


src/settingsSync.ts
```diff
@@ -16,7 +35,75 @@ interface ConfigurationSyncStore {
...
+	return false;
+}
+
+export function parseSyncData(content: string): ISyncData | undefined {
+	try {
+		const syncData: ISyncData = JSON.parse(content);
+		if (isSyncData(syncData)) {
+			return syncData;
+		}
+	} catch {
```
mustard-mh, line 92: should we log like source did?

mustard-mh:
Code looks good 👍 

But I got this error message when I'm testing it

image

akosyakov:
Please make sure to test other scenarios like not settings sync enabled, enabled with GH instead. How does it work then?

src/remoteConnector.ts
```diff
@@ -900,12 +903,76 @@ export default class RemoteConnector extends Disposable {
...
+		const syncDataContent = parseSyncData(syncData.content);
+		if (!syncDataContent) {
+			this.logger.error('Error while parsing sync data');
+			return;
+		}
+
+		let extensions: ISyncExtension[];
+		try {
+			extensions = JSON.parse(syncDataContent.content);
+		} catch {
```
akosyakov, line 958: i think it is important always log underlying issues, please don't skip error

src/remoteConnector.ts
```diff
@@ -900,12 +903,76 @@ export default class RemoteConnector extends Disposable {
...
+			this.logger.error(`Could not execute '__gitpod.cancelGitpodRemoteHeartbeat' command`);
+			this.telemetry.sendTelemetryEvent('vscode_desktop_heartbeat_state', { enabled: String(false), gitpodHost: connectionInfo.gitpodHost, workspaceId: connectionInfo.workspaceId, instanceId: connectionInfo.instanceId, gitpodVersion: gitpodVersion.raw });
+		}
+	}
+
+	private async initializeRemoteExtensions() {
+		let syncData: { ref: string; content: string } | undefined;
+		try {
+			syncData = await this.settingsSync.readResource(SyncResource.Extensions);
```
akosyakov, line 929: I wonder should not we rather provider gitpodHost here from connectionInfo instead of reading from settings. I think for remote window connectionInfo is source of the truth.

mustard-mh:
> not settings sync enabled

It will ask to enable it (with notification and enable action)

> enabled with GH instead

Same as above

mustard-mh:
Is it better to show the process bar when installing the extensions in remote? @jeanp413

src/settingsSync.ts
```diff
@@ -95,22 +185,82 @@ export default class SettingsSync extends Disposable {
...
+		return `${new URL(serviceUrl).toString().replace(/\/$/, '')}/code-sync`;
+	}
+
+	public async readResource(path: string) {
+		if (!this.isSyncStoreConfigured()) {
+			throw new NoSyncStoreError();
+		}
+
+		const syncStoreURL = this.getServiceUrl();
+		const resourceURL = `${syncStoreURL}/v1/resource/${path}/latest`;
```
mustard-mh, line 215: Note source here https://github.com/gitpod-io/gitpod/blob/main/components/server/src/code-sync/code-sync-service.ts

src/settingsSync.ts
jeanp413: I'll add a log but I'd expect this to always work, if for some reason it failed then it would be mean there's a bigger issue as that would mean settings sync is broken in vscode web too

jeanp413: Ah this is logged in `remoteConnectort.ts` line 951, as it should never be undefined

src/remoteConnector.ts
jeanp413: There's already a log for this

jeanp413:
Added a notification with progress

mustard-mh:
It works well, and show logs really helpful 😄

Maybe we can also increase the timeout of requests to `5000ms`? `1500ms` is too short for me (today's network 🙈)

mustard-mh (approved):

mustard-mh:
Also wandering the release process, will we wait until `__gitpod.initializeRemoteExtensions` command is registered in backend?

akosyakov:
Should we enable it for all users? Can it break a user flow somehow? Should we go via experiment and analytics? i.e. land this PR only for prerelease for now, then start working on analytics and experiment? I would say at least without meaningful analytics we should not put in stable. cc @loujaybee

jeanp413:
Increased the timeout and added `gitpod.remote.syncExtensions` setting will be true by default in pre-release and false in stable

jeanp413:
Merging this and will publish a pre-release version

src/remoteConnector.ts
akosyakov: We were discussing yesterday with self hosting team that having different configurations per host will be good, i.e. one should be able to specify different values for different hosts. Like use SSH Gateway for gitpod.io, but local app for SH solution and so on. How we designed it with one host does not really fit to it.

PR #20, opened 2022-08-31
Ask for local app fallback when `Timed out while waiting for handshake`

jeanp413:
## Description

Ask for local app fallback when `Timed out while waiting for handshake`
Add `gitpodHost` custom attribute to configcat

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12475

## How to test
1. Block ssh connections and check a notification is shown, Clicking Ok will fallback to use local-app

To block port 22 you can follow https://www.digitalocean.com/community/tutorials/how-to-set-up-a-firewall-with-ufw-on-ubuntu-18-04 if you are on linux

### DIFF
--- a/src/remoteConnector.ts
+++ b/src/remoteConnector.ts
@@ -565,6 +565,7 @@ export default class RemoteConnector extends Disposable {
 			}).connect({
 				host: sshDestInfo.hostName,
 				username: sshDestInfo.user,
+				readyTimeout: 40000,
 				authHandler(_methodsLeft, _partialSuccess, _callback) {
 					return {
 						type: 'password',
@@ -786,7 +787,7 @@ export default class RemoteConnector extends Disposable {
 
 		// Only use experiment for SaaS
 		const forceUseLocalApp = getServiceURL(params.gitpodHost) === 'https://gitpod.io'
-			? (await this.experiments.get<boolean>('gitpod.remote.useLocalApp', session.account.id))!
+			? (await this.experiments.get<boolean>('gitpod.remote.useLocalApp', session.account.id, { gitpodHost: params.gitpodHost }))!
 			: vscode.workspace.getConfiguration('gitpod').get<boolean>('remote.useLocalApp')!;
 		const userOverride = isUserOverrideSetting('gitpod.remote.useLocalApp');
 		let sshDestination: string | undefined;
@@ -806,8 +807,22 @@ export default class RemoteConnector extends Disposable {
 				this.telemetry.sendRawTelemetryEvent('vscode_desktop_ssh', { kind: 'gateway', status: 'failed', reason: e.toString(), ...params, gitpodVersion: gitpodVersion.raw, userOverride });
 				if (e instanceof NoSSHGatewayError) {
 					this.logger.error('No SSH gateway:', e);
-					vscode.window.showWarningMessage(`${e.host} does not support [direct SSH access](https://github.com/gitpod-io/gitpod/blob/main/install/installer/docs/workspace-ssh-access.md), connecting via the deprecated SSH tunnel over WebSocket.`);
-					// Do nothing and continue execution
+					const ok = 'Ok';
+					const action = await vscode.window.showWarningMessage(`${e.host} does not support [direct SSH access](https://github.com/gitpod-io/gitpod/blob/main/install/installer/docs/workspace-ssh-access.md), connecting via the deprecated SSH tunnel over WebSocket.`, ok);
+					if (action === ok) {
+						// Do nothing and continue execution
+					} else {
+						return;
+					}
+				} else if (e instanceof SSHError && e.message === 'Timed out while waiting for handshake') {
+					this.logger.error('SSH test connection error:', e);
+					const ok = 'Ok';
+					const action = await vscode.window.showWarningMessage(`Timed out while waiting for SSH handshake, it's possible SSH connections on port 22 are blocked or your network is too slow, connecting via the deprecated SSH tunnel over WebSocket.`, ok);
+					if (action === ok) {
+						// Do nothing and continue execution
+					} else {
+						return;
+					}
 				} else if (e instanceof NoRunningInstanceError) {
 					this.logger.error('No Running instance:', e);
 					vscode.window.showErrorMessage(`Failed to connect to ${e.workspaceId} Gitpod workspace: workspace not running`);
@@ -934,8 +949,8 @@ export default class RemoteConnector extends Disposable {
 				const action = await vscode.window.showInformationMessage(`Could not install local extensions on remote workspace, please enable Settings Sync with Gitpod.`, addSyncProvider, config);
 				if (action === addSyncProvider) {
 					vscode.commands.executeCommand('gitpod.syncProvider.add');
-				} else if(action === config) {
-					vscode.commands.executeCommand('workbench.action.openSettings',`@ext:${this.context.extension.id} sync extensions`);
+				} else if (action === config) {
+					vscode.commands.executeCommand('workbench.action.openSettings', `@ext:${this.context.extension.id} sync extensions`);
 				}
 			} else if (e instanceof NoSettingsSyncSession) {
 				const enableSettingsSync = 'Enable Settings Sync';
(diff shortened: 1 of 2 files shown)


src/remoteConnector.ts
```diff
@@ -806,8 +807,22 @@ export default class RemoteConnector extends Disposable {
...
+					const action = await vscode.window.showWarningMessage(`${e.host} does not support [direct SSH access](https://github.com/gitpod-io/gitpod/blob/main/install/installer/docs/workspace-ssh-access.md), connecting via the deprecated SSH tunnel over WebSocket.`, ok);
+					if (action === ok) {
+						// Do nothing and continue execution
+					} else {
+						return;
+					}
+				} else if (e instanceof SSHError && e.message === 'Timed out while waiting for handshake') {
+					this.logger.error('SSH test connection error:', e);
+					const ok = 'Ok';
+					const action = await vscode.window.showWarningMessage(`Timed out while waiting for SSH handshake, it's possible SSH connections on port 22 are blocked or your network is too slow, connecting via the deprecated SSH tunnel over WebSocket.`, ok);
```
jeanp413, line 820: cc @akosyakov if you want to tweak the message

akosyakov: it looks good to me

cc @loujaybee

src/remoteConnector.ts
```diff
@@ -565,6 +565,7 @@ export default class RemoteConnector extends Disposable {
 			}).connect({
 				host: sshDestInfo.hostName,
 				username: sshDestInfo.user,
+				readyTimeout: 40000,
```
akosyakov, line 568: What is it seconds?

akosyakov:
> Block ssh connections

@jeanp413 Any guidance how to do it? Maybe we should document it in https://www.notion.so/gitpod/VS-Code-Knowledge-sharing-FAQ-s-23145c3d43dd4b6db25b9f02486de3e0

src/remoteConnector.ts
filiptronicek, line 820: @jeanp413 I'd personally suggest breaking it into multiple sentences like this:

```suggestion
					const action = await vscode.window.showWarningMessage(`Timed out while waiting for the SSH handshake. It's possible, that SSH connections on port 22 are blocked, or your network is too slow. Connecting via the deprecated SSH tunnel over WebSocket instead.`, ok);
```

src/remoteConnector.ts
```diff
@@ -806,8 +807,22 @@ export default class RemoteConnector extends Disposable {
 				this.telemetry.sendRawTelemetryEvent('vscode_desktop_ssh', { kind: 'gateway', status: 'failed', reason: e.toString(), ...params, gitpodVersion: gitpodVersion.raw, userOverride });
 				if (e instanceof NoSSHGatewayError) {
 					this.logger.error('No SSH gateway:', e);
-					vscode.window.showWarningMessage(`${e.host} does not support [direct SSH access](https://github.com/gitpod-io/gitpod/blob/main/install/installer/docs/workspace-ssh-access.md), connecting via the deprecated SSH tunnel over WebSocket.`);
-					// Do nothing and continue execution
+					const ok = 'Ok';
```
filiptronicek, line 810: nit: change casing [[1](https://www.writing-skills.com/ok-ok-okay-how-do-you-write-ok)]
```suggestion
					const ok = 'OK';
```

filiptronicek: Actually @jeanp413 can we define this variable a level up so that we don't have to initialize the same variable on line 819?

jeanp413: I don't want to pollute other scopes as the actions are different depending on message, if this increases then I'll create another file with actions constants :thinking:

src/remoteConnector.ts
jeanp413: from @filiptronicek https://github.com/gitpod-io/gitpod-vscode-desktop/pull/20#discussion_r959299375

jeanp413:
> > Block ssh connections
> 
> @jeanp413 Any guidance how to do it? Maybe we should document it in https://www.notion.so/gitpod/VS-Code-Knowledge-sharing-FAQ-s-23145c3d43dd4b6db25b9f02486de3e0

Added a link in the description for linux, don't know if it works for mac

src/remoteConnector.ts
```diff
@@ -806,8 +807,22 @@ export default class RemoteConnector extends Disposable {
...
+					const action = await vscode.window.showWarningMessage(`${e.host} does not support [direct SSH access](https://github.com/gitpod-io/gitpod/blob/main/install/installer/docs/workspace-ssh-access.md), connecting via the deprecated SSH tunnel over WebSocket.`, ok);
+					if (action === ok) {
```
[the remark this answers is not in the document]
mustard-mh: Maybe no action buttons? Just let users know that we're going to connect via `local-app`.

If we connect via `local-app` only when user clicks `ok`, it should be asking, I mean end with `?`

src/remoteConnector.ts
mustard-mh: Is it possible to show a process bar when connecting and show "fallback to/try with local-app" action after 20 seconds? 40s or 1-2minutes are too long to wait

src/remoteConnector.ts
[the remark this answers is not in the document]
jeanp413: > Maybe no action buttons? Just let users know that we're going to connect via local-app.

That's how it was until this PR :laughing:

src/remoteConnector.ts
jeanp413: > Is it possible to show a process bar when connecting

yeah possible but this connection does nothing it's just a quick test to check if port is not blocked so not sure if we should expose it to the user

src/remoteConnector.ts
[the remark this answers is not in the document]
jeanp413: > How does UI look like? 2 buttons Ok and Cancel? Can we have only one Ok?

It's only one button right now but the user can click on the `x` and that would return `undefined`

src/remoteConnector.ts
akosyakov: > Is it possible to show a process bar when connecting and show "fallback to/try with local-app" action after 20 seconds? 40s or 1-2minutes are too long to wait

I mean generally it should be very quick, so maybe it is not bad to show progress bar if you really have problems, then we need to add some messaging `Verifying SSH connection...`

src/remoteConnector.ts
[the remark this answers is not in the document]
akosyakov: ok, UI wise it is alright, but clicking on `x` or pressing Esc should be the same like  `Ok`. i.e. we should lead a user to success even if it is with local app for now.

akosyakov (approved):
I think code-wise it looks good beside: https://github.com/gitpod-io/gitpod-vscode-desktop/pull/20/files#r960355213

we should always continue execution does not matter how a user closed the warning with cancel or OK

I have not tried it. @mustard-mh @filiptronicek could you help with testing 🙏

mustard-mh:
Tested by change source code

- NoSSHGatewayError - change `sshHostKeyEndPoint` requests URL
- Timeout - change `readyTimeout` to 1 ms

All good, as close notification will also fallback to local-app [[1](https://github.com/gitpod-io/gitpod-vscode-desktop/pull/20#discussion_r960355213)]

mustard-mh (approved):

src/remoteConnector.ts
```diff
@@ -786,7 +787,7 @@ export default class RemoteConnector extends Disposable {
 
 		// Only use experiment for SaaS
 		const forceUseLocalApp = getServiceURL(params.gitpodHost) === 'https://gitpod.io'
-			? (await this.experiments.get<boolean>('gitpod.remote.useLocalApp', session.account.id))!
+			? (await this.experiments.get<boolean>('gitpod.remote.useLocalApp', session.account.id, { gitpodHost: params.gitpodHost }))!
```
mustard-mh, line 790: This `gitpodHost` is always `https://gitpod.io`

src/remoteConnector.ts
iQQBot: Yes, this test should be very quick, if handshake take 40s, It's hard to imagine what the latency would be once you've successfully connected

src/remoteConnector.ts
akosyakov: for now, later we are going to remove `?`, but first we should release this PR to both stable and latest and see effect on gitpod.io

PR #21, opened 2022-09-06
flow analytics against vscode apis

akosyakov:
## Description

This PR normalize analytics events and prepares an experiment for synching local extensions: 
- all events start with `vscode_desktop_` prefix to distinguish from other
- events are instrumented with information about host, user and workspace
- events have `flow` and `status` properties, like `ssh`, `settings-sync`, `sync_local_extensions` flows
- VS Code notifications contribute to flows automatically by instrumenting VS Code APIs

## Related Issue(s)

Fixes #

## How to test

Debug extension, check out analytics in the lab project while connecting via remote ssh and synching local extensions.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/src/authentication.ts
+++ b/src/authentication.ts
@@ -12,6 +12,8 @@ import Log from './common/logger';
 import { arrayEquals } from './common/utils';
 import { Disposable } from './common/dispose';
 import TelemetryReporter from './telemetryReporter';
+import { UserFlowTelemetry } from './common/telemetry';
+import { NotificationService } from './notification';
 
 interface SessionData {
 	id: string;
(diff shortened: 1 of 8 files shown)


akosyakov:
@jeanp413 @mustard-mh could you have a look please? If everything is alright feel free to deploy to latest and stable.

Do I understand correctly that if ConfigCat does not have defined experiment then we automatically falls back to defaults? Or I should define an experiment first for production?

src/vscodeTelemetry.ts
```diff
@@ -0,0 +1,58 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import TelemetryReporter from './telemetryReporter';
+
+/**
+ * Intruments telemetry for VS Code Extension API triggering or triggered by user actions.
+ */
+export function instrumentVSCodeTelemetry(telemetry: TelemetryReporter): void {
```
jeanp413, line 12: Thinking more about this, this looks too hacky :smile: , how about having a `notificationService` class instead that just wraps the vscode api, adds the telemetry stuff and use that instead in all places

akosyakov: let's do it

akosyakov:
@jeanp413 I've refactored to use a dedicated service. Could you review please?

If everything is alright, let's merge and start an experiment next week. I will also update all events in tracking plan that they start now with `vscode_desktop_` and has properties like `status`, `gitpodHost` and so on.

jeanp413 (approved):

PR #22, opened 2022-09-06
Add PR CI workflow

filiptronicek:
## Description
Adds a GitHub Action workflow to build and package the extension for simple CI.

## How to test
As it goes with GH Actions, other than code review, merge and see 😆.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,44 @@
+name: Build
+
+on:
+  pull_request:
+
+jobs:
+  nightly:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - name: Install dependencies
+        run: |
+          yarn install --frozen-lockfile
+
+      - name: Update package.json
+        run: |
+          set -e
+          setSegmentKey="setpath([\"segmentKey\"]; \"${{ secrets.ANALITYCS_KEY }}\")"
+          setConfigcatKey="setpath([\"configcatKey\"]; \"${{ secrets.CONFIGCAT_KEY }}\")"
+          jqCommands="${setSegmentKey} | ${setConfigcatKey}"
+          cat package.json | jq "${jqCommands}" > package.json.tmp
+          mv package.json.tmp package.json
+
+      - name: Generate package.json
+        run: |
+          set -e
+          node ./scripts/prepare-nightly-build.js
+          mv package.insiders.json package.json
+
+      - name: Package extension
+        id: package_vsix
+        run: |
+          yarn package --pre-release
+          echo ::set-output name=vsix_path::$(ls *.vsix)
+
+      - name: Upload built extension
+        uses: actions/upload-artifact@v3
+        with:
+          name: extension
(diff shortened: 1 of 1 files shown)


akosyakov (approved):
lgtm

.github/workflows/pr.yml
```diff
@@ -0,0 +1,44 @@
...
+          node-version: 16
+
+      - name: Install dependencies
+        run: |
+          yarn install --frozen-lockfile
+
+      - name: Update package.json
+        run: |
+          set -e
+          setSegmentKey="setpath([\"segmentKey\"]; \"${{ secrets.ANALITYCS_KEY }}\")"
```
akosyakov, line 22: we should not set keys from secrets, it should be used only for (pre)-releases

@filiptronicek could you remove it?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,44 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - name: Install dependencies
+        run: |
+          yarn install --frozen-lockfile
```
jeanp413, line 17: Add nocde_modules caching here see openvscode-releases github action

filiptronicek: Added, thanks!

.github/workflows/pr.yml
filiptronicek: > we should not set keys from secrets, it should be used only for (pre)-releases

Wondered whether that was a good idea, removed it now, thanks for the input!

@akosyakov are you opposed to having artifacts in the action's outputs? I would find it very helpful so that a) as the PR creator I don't have to build the extension and upload to the PR's description and b) if I want to review a PR, I can always get the latest artifact without having to build myself, which eliminates me having to pull the latest changes, installing dependencies and packaging up the extension.

Agreed that having `yarn build` is very much enough for just CI, but to help the whole process these packaged artifacts would be cool 😄.

jeanp413: We no longer upload vsix for this extension, we did it before because it was inside vscode repo, we still do this for gitpod-remote only.
If someone wants to review then he needs to clone the repo and run locally, also for testing telemetry and check them in segment by default it queues events so usually you want to change that behavior manually in the code

filiptronicek: Thanks for clearing this up, @jeanp413 🙏. Removed this in https://github.com/gitpod-io/gitpod-vscode-desktop/pull/22/commits/d80cdb4eefd635a20de049a4f9b92e2f15002892.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,22 @@
...
+          node-version: 16
+          cache: "yarn"
+
+      - name: Install dependencies
+        run: |
+          yarn install --frozen-lockfile
+
+      - name: Package extension
+        run: |
+          yarn package
```
jeanp413, line 22: use `yarn compile` instead?

filiptronicek: It's important to ensure that we can package the extension as well, don't you think? 👀

Fine with removing this, but I feel this adds an easy way to catch some issues with packaging and it shouldn't affect checking transpiling in any way.

PR #23, opened 2022-09-08
[config-cat] refresh settings values once in a minute on demand

akosyakov:
## Description

With this PR we refresh ConfigCat settings values only once in a minute on demand.

You can test by using `Developer: Refresh Gitpod Experiments` command and observing last refresh date and time diff relative to it. If diff is over 60000 milliseconds new refresh happens. This command is available only while running from sources. The logic should work across multiple windows, so if you restart the debug session you should see that previous should be respected.

Screenshot 2022-09-08 at 09 59 03

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,12 @@
 				"command": "gitpod.showReleaseNotes",
 				"category": "Gitpod",
 				"title": "Show Release Notes"
+			},
+			{
+				"command": "gitpod.dev.refreshExperiments",
+				"category": "Developer",
+				"title": "Refresh Gitpod Experiments",
+				"enablement": "gitpod.dev"
 			}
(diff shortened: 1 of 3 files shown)


jeanp413:
Why are we changing this? `createClientWithLazyLoad` is better, it only requests config.json when user does an action that requires experiments and that's only when opening a workspace which is way less than fetching it every minute, do we have a metric how many workplaces get's opened in vscode desktop per user in a  day?

akosyakov:
Oh, I thought it is lazy only first time then it fetches each minute? If it is not true and cache only invalidated on demand then it is good already 👍🏻

PR #24, opened 2022-09-16
Improve ports

jeanp413:
## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8377

## How to test

1. Open a workspace with ports configured in gitpod.yml
2. Check the ports are forwarded when connecting in ports view
3. Run some command that uses some port
4. Check the port is forwarded automatically

## Release Notes

```release-note
NONE
```

### DIFF
--- a/src/remoteConnector.ts
+++ b/src/remoteConnector.ts
@@ -5,6 +5,7 @@
 
 import { AutoTunnelRequest, ResolveSSHConnectionRequest, ResolveSSHConnectionResponse } from '@gitpod/local-app-api-grpcweb/lib/localapp_pb';
 import { LocalAppClient } from '@gitpod/local-app-api-grpcweb/lib/localapp_pb_service';
+import { WorkspaceConfig, PortConfig, PortOnOpen } from '@gitpod/gitpod-protocol/lib/protocol';
 import { NodeHttpTransport } from '@improbable-eng/grpc-web-node-http-transport';
 import { grpc } from '@improbable-eng/grpc-web';
 import * as cp from 'child_process';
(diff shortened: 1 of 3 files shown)


src/remoteConnector.ts
```diff
@@ -1028,6 +1030,71 @@ export default class RemoteConnector extends Disposable {
 		}
 	}
 
+	private async configureMachineSettings(wsConfig: WorkspaceConfig) {
```
akosyakov, line 1033: Should not this logic implemented in remote extension? It has access to machine setitngs, supervisor and VS Code api. I don't see a point to implement in local extension.

src/remoteConnector.ts
```diff
@@ -1057,6 +1124,8 @@ export default class RemoteConnector extends Disposable {
 
 		await this.context.globalState.update(`${RemoteConnector.SSH_DEST_KEY}${sshDestStr}`, { ...connectionInfo, isFirstConnection: false });
 
+		this.configureMachineSettings(workspaceInfo.workspace.config).then(() => this.tunnelPortsFromConfig(workspaceInfo.workspace.config.ports ?? []));
```
akosyakov, line 1127: config.ports are not real ports, they are intitial configuration for supervisor.

We should observe port status endpoint from supervisor and reconcile it with VS Code as here: https://github.com/gitpod-io/openvscode-server/blob/849af772ba8e6a3a2944ccaf004d70cb549d3f3e/extensions/gitpod-web/src/extension.ts#L551-L592

src/remoteConnector.ts
```diff
@@ -1028,6 +1030,71 @@ export default class RemoteConnector extends Disposable {
...
+
+		if (!extRemoteLogsUri) {
+			return;
+		}
+
+		const remoteUserDataPath = path.posix.dirname(path.posix.dirname(path.posix.dirname(path.posix.dirname(extRemoteLogsUri.path))));
+		const machineSettingsResource = extRemoteLogsUri.with({ path: path.posix.join(remoteUserDataPath, 'Machine', 'settings.json') });
+		try {
+			let settingsStr: string = '{}';
+			const fileExists = await vscode.workspace.fs.stat(machineSettingsResource).then(() => true, () => false);
```
akosyakov, line 1051: Should not we use document API to respect dirty changes from a user?

PR #25, opened 2022-10-17
Test public API

jeanp413:
## Description
Related https://github.com/gitpod-io/gitpod/pull/13897

## How to test

Check workspace opens successfully


src/publicApi.ts
```diff
@@ -0,0 +1,48 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { createConnectTransport, createPromiseClient, Interceptor, PromiseClient } from '@bufbuild/connect-web';
+import { WorkspacesService } from '@gitpod/public-api/lib/gitpod/v1/workspaces_connectweb';
+import { Workspace, WorkspaceInstance } from '@gitpod/public-api/lib/gitpod/v1/workspaces_pb';
+
+export class GitpodPublicApi {
+
+    private workspaceService!: PromiseClient<typeof WorkspacesService>;
```
mustard-mh, line 12: Why not make it public and put all needed services here with init or cache (cache with a getter?)

mustard-mh:
It should work, but it's not working now.

Maybe because we use the new version of client (move to `experimental` folder) but server not been deployed yet

- Server version of SaaS Gitpod `main.5223`
- Package version `0.1.5-main.5229`
- Milan's [changes](https://github.com/gitpod-io/gitpod/commit/2d888a5ded8f31f855e867b1655b480c527d0329) `main.5229`

So let's wait until tomorrow WebApp deployment to test this PR

jeanp413:
> Will we bring this PR to production? 

yes that's my understanding cc @akosyakov 

> It should work, but it's not working now (API respond with 404)

yeah that's fine, let's wait for webapp deployment

akosyakov:
yes, we should bring it in production, but enable only for some users to test, most of users should still use Server API and MUST NOT be broken.

mustard-mh (approved):
Workspace opens successfully 🎉 

Tested with both public/server API workspaces open and heartbeat

jeanp413:
ship it then :skateboard:

jeanp413:
for some reason webpack doesn't like public api :facepalm: 
https://github.com/gitpod-io/gitpod-vscode-desktop/actions/runs/3283268902/jobs/5407774187

jeanp413:
This will fix webpack build https://github.com/gitpod-io/gitpod/pull/14058

PR #27, opened 2022-11-29
Remove installing extensions notification

filiptronicek:
Removes this:
image

because of [this](https://gitpod.slack.com/archives/C045J13P8A1/p1669719254002629) (Slack message)


akosyakov (approved):
code looks good, but I did not try

PR #31, opened 2023-02-06
Force Linux host platform for Windows users

filiptronicek:
## Description
Force VS Code to think the server it's connecting to is a Linux host, which caused detection problems on Windows. 

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/16058

## How to test
1. Connect to a workspace with VS Code Desktop (if you can, go get a Windows machine for this)
2. Don't get the OS prompt


filiptronicek:
It looks like it still prompts on Windows, although being in the config :/ 
image

jeanp413 (approved):

PR #32, opened 2023-02-10
Add restart dialog

jeanp413:
## Description

When workspace stops, the extension detects it and then shows a dialog with an option to open the workspace url in the browser so the user can restart it.

## How to test

1. Install this vsix [gitpod-desktop-0.0.93.zip](https://github.com/gitpod-io/gitpod-vscode-desktop/files/10814938/gitpod-desktop-0.0.93.zip)
2. Create workspace and open it in vscode desktop

3. Stop workspace
4. A dialog will show telling the user the workspace has stopped with an option to restart the workspace

~~Observation:~~
~~- The restart dialog will race with the remote-ssh connection lost dialog, so could show first or after~~


jeanp413:
Recording 2023-02-09 at 22 37 21

jeanp413:
Couldn't get the `streamWorkspaceStatus` api to work, it works for a couple minutes but then hangs indefinitely after 5 min or so, it doesn't throw any error :disappointed: , maybe a bug in the connect library? cc @akosyakov

akosyakov:
Can we try to close the window automatically as well?

akosyakov:
@jeanp413 maybe check with Milan about the connect lib. It should work some how. I think Huiwen and Pudong can also help to investigate

src/workspaceState.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        this.logger.trace(`WorkspaceState manager for workspace ${workspaceId} started`);
+
+        // this._register(this.publicApi.onWorkspaceStatusUpdate(u => this.checkWorkspaceState(u)));
+        // this.publicApi.startWorkspaceStatusStreaming(workspaceId);
+
+        this.pollWorkspaceState();
+        this.poolHandle = setInterval(() => {
+            this.pollWorkspaceState();
+        }, WorkspaceState.POLL_INTERVAL);
```
filiptronicek, line 41: question: is it correct that here it can happen that before we can display our workspace stopping dialog, the VS Code SSH dialog will show first?

jeanp413: yes, if remote-ssh extension detects connection is lost first it will show before our dialog

src/remoteConnector.ts
```diff
@@ -998,14 +1011,28 @@ export default class RemoteConnector extends Disposable {
...
+		const msg = workspaceUrl
+			? `Workspace ${workspaceId} is not running. Please restart the workspace.`
+			: `Workspace not found. Please start the workspace from dashboard.`;
+		this.logger.error(msg);
+
+		const openUrl = 'Restart workspace';
+		const resp = await this.notifications.showErrorMessage(msg, { modal: true, id: uuid(), flow }, openUrl);
+		if (resp === openUrl) {
+			await vscode.env.openExternal(vscode.Uri.parse(workspaceUrl || 'https://gitpod.io/workspaces'));
+			vscode.commands.executeCommand('workbench.action.closeWindow');
```
jeanp413, line 1024: > Can we try to close the window automatically as well?

@akosyakov  It closes automatically after clicking on open button

jeanp413:
Update:

restartDemo

PR #36, opened 2023-03-06
Implement local ssh daemon

mustard-mh:
## Description

Implement local ssh daemon

## Related Issue(s)

Related https://github.com/gitpod-io/gitpod/issues/16836
Related https://linear.app/gitpod/issue/IDE-29/spike-compare-localssh-%E2%9A%94%EF%B8%8F-directssh-localapp

## Analysis Path [^1]
- Event vscode_desktop_ssh, with property `kind: local-ssh`
- With funnel: `connecting` -> `? local-ssh-failure` -> connected (`got_destination`)

[^1]: https://github.com/gitpod-io/gitpod-vscode-desktop/pull/36#discussion_r1152898622

## Error Reporting [^1]

Images below are local testing result, it will send to GCP in production
|String|Error|
|:-:|:-:|
|image|image|

## How to test

Should test with platform `stable`, `insiders` and os `linux` `macOS` `Windows` (mostly test with file system)

- Clone this repo and switch to PR's branch
- Exec commands below to install extension locally
```
yarn
yarn local-test
```
- Restart VSCode to make sure daemon is up-to-date (need 10 seconds to exit) [^2], commands below can help test [^3]
- Start workspace in Gitpod and Open in VSCode
- It should able to start without password required
- Stop workspace, and use VSCode Desktop `recent` (ctrl + R) to access the stopped workspace
- It should ask you if you want to restart it
- Run command `Gitpod: Export all Logs` should include local ssh loggings
  - [x] Stable
    - [x] macOS
    - [x] Windows
    - [x] Linux
  - [x] Insiders
    - [x] macOS
    - [x] Windows
    - [x] Linux

### ‼️  Major Test Points ‼️
- [x] ‼️ Local SSH can be controlled by  feature flag https://app.configcat.com/08da1258-64fb-4a8e-8a1e-51de773884f6/08da1258-6541-4fc7-8b61-c8b47f82f3a0/08da1258-6512-4ec0-80a3-3f6aa301f853?settingId=75503
- [ ] Throw error inside localssh (i.e. reject [here](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/f2d52c8e5ef97ef5d7c58d4168413320bcd825c1/src/local-ssh/sshTunnel.ts#L59)) errors should reported to segment https://app.segment.com/gitpod/sources/ide_team_playground/debugger
- [ ] Should send event to segment https://app.segment.com/gitpod/sources/ide_team_playground/debugger with `kind = local-ssh`
- [x] ‼️ It can connect with LocalSSH
  - [x] Stable
    - [x] macOS
    - [x] Windows
    - [x] Linux
  - [x] Insiders
    - [x] macOS
    - [x] Windows
    - [x] Linux

[^2]: We need 10 second to avoid daemon exists too quickly with only one window trying to reload
[^3]: watch daemon process `watch -n 1 "ps aux | grep '[d]aemon.js'"`  and tail daemon loggings `tail -f $TMPDIR/gitpod-vscode-daemon.log | grep <pid>`


src/local-ssh/common.ts
```diff
@@ -0,0 +1,97 @@
...
+		this.logger.warn(message, ...args);
+	}
+	error(error: string | Error, ...args: any[]): void {
+		this.logger.error(error as any, ...args);
+	}
+}
+
+function getIPCHandlePath(id: string, isAddr: boolean = false): string {
+	if (process.platform === 'win32') {
+		return `\\\\.\\pipe\\gp-${id}-sock`;
```
akosyakov, line 72: let's use something more specific like `gitpod-vscode-`

src/local-ssh/ipc/extension.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        this.backoffActive();
+
+        setTimeout(() => this.pingLocalSSHService(), 1000);
+
+        // TODO(hw): ping local ssh service to make sure it's alive
+        // if not, restart it
+    }
+
+    private async backoffActive() {
```
akosyakov, line 113: we should report it to GCP if it fails

akosyakov: also would be good to have analytics whether it was activated

src/local-ssh/ipc/extension.ts
```diff
@@ -0,0 +1,141 @@
...
+        const serviceImpl = new ExtensionServiceImpl(this.logService, this.sessionService, this.hostService, this.notificationService);
+        server.add(ExtensionServiceDefinition, serviceImpl);
+        server.listen(getExtensionIPCHandleAddr(this.id)).then(() => {
+            this.logService.info('extension ipc service server started to listen with id: ' + this.id);
+        }).catch(this.logService.error);
+        this.server = server;
+
+        this.backoffActive();
+
+        setTimeout(() => this.pingLocalSSHService(), 1000);
```
akosyakov, line 107: should it be it interval?

src/daemonStarter.ts
```diff
@@ -0,0 +1,40 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { join } from 'path';
+import { spawn } from 'child_process';
+import { ExitCode } from './local-ssh/common';
+import { ILogService } from './services/logService';
+
+const sleep = (sec: number) => new Promise(resolve => setTimeout(resolve, sec * 1000));
+
+export async function ensureDaemonStarted(logService: ILogService, retry = 10) {
```
akosyakov, line 13: Should not it go under local-ssh?

src/daemonStarter.ts
```diff
@@ -0,0 +1,40 @@
...
+        ensureDaemonStarted(logService, retry - 1);
+    });
+    await sleep(1);
+}
+
+export async function tryStartDaemon(logService: ILogService) {
+    logService.info('going to start local-ssh daemon');
+    const daemon = spawn(process.execPath, [join(__dirname, 'local-ssh/daemon.js')], {
+        detached: true,
+        stdio: 'ignore',
```
akosyakov, line 35: where logs goes? It woudl be good if Export Logs command can capture it and a user coudl understand from `Gitpod` output view what is happening.

src/daemonStarter.ts
```diff
@@ -0,0 +1,40 @@
...
+                logService.error('exit with code: ' + code);
+                return;
+        }
+        logService.error('unexpectedly exit with code: ' + code + ' attempt retry: ' + retry);
+        ensureDaemonStarted(logService, retry - 1);
+    });
+    await sleep(1);
+}
+
+export async function tryStartDaemon(logService: ILogService) {
```
akosyakov, line 31: if it fails because it is not already running then we should log an error to GCP, we should have an analytics event for whether it is succesfully started

src/local-ssh/sshTunnel.ts
```diff
@@ -0,0 +1,112 @@
...
+
+	private get workspaceUrl() {
+		return `https://${this.workspaceInfo.workspaceId}.${this.workspaceInfo.workspaceHost}`;
+	}
+
+	private get workspaceWSUrl() {
+		return `wss://${this.workspaceInfo.workspaceId}.${this.workspaceInfo.workspaceHost}`;
+	}
+
+	public async establishTunnel(): Promise<ConnectConfig> {
```
akosyakov, line 76: Could you map out analytics funnel how we will track conversion from hadnling uri to getting the session please. We should make sure to install all necessary analytics events and put GCP error reporting in such points.

src/local-ssh/common.ts
```diff
@@ -0,0 +1,97 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import winston from 'winston';
+import { existsSync, unlinkSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { GetWorkspaceAuthInfoResponse } from '../proto/typescript/ipc/v1/ipc';
+
+export const PID_LOCK_FILE = '/tmp/gp-daemon.lock';
```
akosyakov, line 12: this is unsafe

src/extension.ts
```diff
@@ -107,6 +112,8 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	context.subscriptions.push(new ReleaseNotes(context, commandManager, logger));
 
+	await ensureDaemonStarted(logger);
```
akosyakov, line 115: I wonder should we really do it here? It seems we should retry it sometimes based on events or something from ExtensionServiceServer?

akosyakov (commented):
Great work, there are many things to improve yet, but I wonder how we can bring it to dogfooding via feature flag.

src/local-ssh/ipc/extension.ts
mustard-mh: We do if for purpose, to avoid we get two `ping` function running. (inside `pingLocalSSHService`, it will call it self again)

src/daemonStarter.ts
mustard-mh: We can't get vscode context inside daemon, even we pipe outputs into vscode outputs in `extension.ts`, the window that start this daemon may be closed.

How about add this `.log` file into `Export Logs` files?

src/local-ssh/common.ts
mustard-mh: I'm going remove this `.lock` file as it's useless, we can check it via local ssh server's port used or not

src/common/telemetry.ts
```diff
@@ -152,6 +152,7 @@ export class BaseTelemetryReporter extends Disposable {
 		private extensionVersion: string,
 		private telemetryAppender: ITelemetryAppender,
 		private osShim: { release: string; platform: string; architecture: string },
+		private isProduction: boolean,
```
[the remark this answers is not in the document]
mustard-mh: If I want to test local-ssh locally, I need to install it. So add this checking to avoid local debug analytics data send to production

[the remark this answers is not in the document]
akosyakov: I don't think you need to pass it here, it can be done in TelemetryService in client factory callback?

src/remoteConnector.ts
```diff
@@ -669,7 +717,7 @@ export class RemoteConnector extends Disposable {
 					await this.showSSHPasswordModal(password, gatewayFlow);
 				}
 
-				this.telemetryService.sendUserFlowStatus('connected', gatewayFlow);
+				this.telemetryService.sendUserFlowStatus('got_destination', gatewayFlow);
```
mustard-mh, line 720: @akosyakov I'd like to change this status to `got_destination` since it's not really connected yet. Good for you?

akosyakov: how the funnel is going to look for all modes?

akosyakov:
Let's make sure that this PR does not break existing semantic, hide local SSH gateway behind feature flag and merge, try to contain changes. No need to look in analytics yet.

src/local-ssh/sshTunnel.ts
mustard-mh: See `Analysis Path` section in PR description

src/remoteConnector.ts
mustard-mh: > how the funnel is going to look for all modes?

What do you mean? 

For other connect ways (`direct ssh gateway` and `local-app`) they should work like before except naming changed to `got_destination`. 

local-ssh funnel will go:
- Event vscode_desktop_ssh, with property `kind: local-ssh`
- With funnel: `connecting` -> `? local-ssh-failure` -> connected (`got_destination`)

src/experiments.ts
```diff
@@ -25,8 +26,9 @@ export class ExperimentalSettings {
         private readonly context: vscode.ExtensionContext,
         private readonly logger: ILogService
     ) {
+        const isProduction = this.context.extension.packageJSON.buildEnv === 'production';
```
akosyakov, line 29: Why do we need another flag in package json? vs code context already have it?

mustard-mh: https://github.com/gitpod-io/gitpod-vscode-desktop/pull/36#discussion_r1168560639

To avoid install extension locally send debugging telemetry to production

src/remoteConnector.ts
akosyakov: let's not do it in this PR

package.json
```diff
@@ -4,6 +4,7 @@
 	"description": "Gitpod Support",
 	"publisher": "gitpod",
 	"version": "0.0.98",
+	"daemonVersion": "0.0.1",
```
akosyakov, line 7: Do we really need 2 versions? How are we going to maintain it?

src/experiments.ts
akosyakov: You can put an env var like `VSCODE_DEV: 1` instead? to avoid messing with package.json during release?

package.json
mustard-mh: That's because our daemon is not always updated when extension upgrade. I want to have notification to ask users to exit vscode desktop to get latest daemon features.

> Do we really need 2 versions?

But yep, we don't need it now

> How are we going to maintain it?

Original thought is to update this field after we change daemon

src/remoteConnector.ts
mustard-mh: > let's not do it in this PR

Revert to `connected` ?

package.json
akosyakov: > Original thought is to update this field after we change daemon

yes, but I'm not sure how ensure it. Anyway I think we are getting to far ahead of us. Let's land it somehow to try internally.

scripts/local-test.js
```diff
@@ -0,0 +1,21 @@
...
+function install() {
+    // bump up
+    const version = new semver.SemVer(packageJson.version)
+    version.inc('patch');
+    packageJson.version = version.version;
+    fs.writeFileSync(path.join(__dirname, '../package.json'), JSON.stringify(packageJson, null, '\t') + "\n");
+
+    // install
+    cp.execSync('yarn package')
+    cp.execSync(`code --install-extension gitpod-desktop-${version.version}.vsix --force`)
```
akosyakov, line 16: but why? we have a debug target? 🤔

src/experiments.ts
mustard-mh: > You can put an env var like VSCODE_DEV: 1 instead? 

Do you mean configure `VSCODE_DEV` in my `.zshrc`? I don't want to mess my local environment.

> to avoid messing with package.json during release?

1️⃣ I added `setBuildEnv="setpath([\"buildEnv\"]; \"production\")"` to `.github/workflow/nightly.yaml` and `.github/workflow/release.yaml` is it enough?

2️⃣ Or I'd like to stash these changes, and apply when I'm going to debug with locally installation.

Which one do you prefer?

scripts/local-test.js
mustard-mh: I need only one daemon (with current code base) to be started. And to make it like the env that user using

package.json
```diff
@@ -137,31 +144,37 @@
...
 	},
 	"devDependencies": {
 		"@types/google-protobuf": "^3.7.4",
 		"@types/js-yaml": "^4.0.5",
 		"@types/mocha": "^9.1.1",
 		"@types/node": "16.x",
 		"@types/semver": "^7.3.10",
-		"@types/ssh2": "^0.5.52",
+		"@types/ssh2": "^1.11.7",
+		"@types/ssh2-streams": "^0.1.9",
```
jeanp413, line 159: revert this as it's unrelated

src/services/telemetryService.ts
jeanp413, line 1: Revert this, I saw anton is doing similar changes about analytics in another PR.
Also not sure why we need this, in development event will be send to playground ide in segment

src/ssh/identityFiles.ts
jeanp413, line 1: revert this, related to unnecessary lib update in package.json

package.json
```diff
@@ -171,21 +188,33 @@
...
 		"node-fetch-commonjs": "^3.2.4",
+		"node-rsa": "^1.1.1",
 		"pkce-challenge": "^3.0.0",
+		"prom-client": "^14.1.1",
+		"protobufjs": "^7.2.2",
 		"semver": "^7.3.7",
 		"ssh-config": "^4.1.6",
-		"ssh2": "^1.10.0",
+		"ssh2": "^1.11.0",
+		"ssh2-streams": "^0.4.10",
```
jeanp413, line 212: revert this too, only the types got reverted

src/local-ssh/ipc/extension.ts
```diff
@@ -0,0 +1,259 @@
...
+    async getWorkspaceAuthInfo(request: GetWorkspaceAuthInfoRequest, _context: CallContext): Promise<{ workspaceId?: string | undefined; workspaceHost?: string | undefined; ownerToken?: string | undefined }> {
+        try {
+            const accessToken = this.sessionService.getGitpodToken();
+            if (!accessToken) {
+                throw new ServerError(Status.INTERNAL, 'no access token found');
+            }
+            const workspaceId = request.workspaceId;
+
+            const gitpodHost = this.hostService.gitpodHost;
+            const usePublicApi = await this.experiments.getUsePublicAPI(gitpodHost);
```
jeanp413, line 69: we agreed to use use public api for new features going forward but not sure which one has more priority to deploy to all users first local ssh or public api 🤔 
@akosyakov should we use just public api for this?

akosyakov: We deploy it only for Gitpodders. Public API should be enabled for them already?

jeanp413: yep, so lets remove the server api logic here @mustard-mh

mustard-mh: > yep, so lets remove the server api logic here @mustard-mh

In this case, for users, it'll only works with public api (no self-hosted support anymore?)

akosyakov: for me it works both ways. I would like to try it with Gitpodders and learn more, we can figure out details later along the road.

tsconfig.json
```diff
@@ -2,6 +2,7 @@
 	"compilerOptions": {
 		"target": "es2020",
```
jeanp413, line 3: change target to `es2021` too

PR #37, opened 2023-03-06
Add grpc metrics

jeanp413:
Adds grpc metrics

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -1,0 +1,378 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Gitpod. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import * as grpc from '@grpc/grpc-js';
+import { Registry, Counter, Histogram, metric } from 'prom-client';
+import { Code, connectErrorFromReason, Interceptor, StreamRequest, UnaryRequest } from '@bufbuild/connect-node';
+import { MethodKind } from '@bufbuild/protobuf';
+import { codeToString, StreamResponse, UnaryResponse } from '@bufbuild/connect-core';
+
+export type GrpcMethodType = 'unary' | 'client_stream' | 'server_stream' | 'bidi_stream';
+
+export interface IGrpcCallMetricsLabels {
+    service: string;
+    method: string;
+    type: GrpcMethodType;
+}
+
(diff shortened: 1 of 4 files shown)


src/metrics.ts
```diff
@@ -0,0 +1,378 @@
...
+    readonly handledSecondsHistogram: Histogram<string>;
+
+    constructor() {
+        this.startedCounter = new Counter({
+            name: 'grpc_client_started_total',
+            help: 'Total number of RPCs started on the client.',
+            labelNames: ['grpc_service', 'grpc_method', 'grpc_type'],
+            registers: [register],
+        });
+        this.sentCounter = new Counter({
```
mustard-mh, line 41: We don't have `grpc_client_msg_sent_total`, `grpc_client_msg_received_total` metrics added in ide-metrics whitelist and don't need them

src/metrics.ts
```diff
@@ -0,0 +1,378 @@
...
+    private async reportCounter(metric: metric) {
+        const counterMetric = metric as metric & { values: [{ value: number; labels: Record<string, string> }] };
+        for (const { value, labels } of counterMetric.values) {
+            if (value > 0) {
+                await this.doAddCounter(counterMetric.name, labels, value);
+            }
+        }
+    }
+
+    private async reportHistogram(metric: metric) {
```
mustard-mh, line 301: For `reportHistogram` in supervisor/reporter.go we need previous value to get diff value. But here is nothing related to previous value, should not we copy logic in supervisor and translate it into js (maybe with the help of copilot lab xd)?

src/metrics.ts
```diff
@@ -0,0 +1,378 @@
...
+            const type = m.type as unknown as string;
+            if (type === 'counter') {
+                await this.reportCounter(m);
+            } else if (type === 'histogram') {
+                await this.reportHistogram(m);
+            }
+        }
+    }
+
+    private async reportCounter(metric: metric) {
```
mustard-mh, line 292: Same as `reportHistogram`

src/metrics.ts
jeanp413: No need to calculate difference as I call `register.resetMetrics();` above, and because this are just counters internally is the same as diff

src/metrics.ts
jeanp413: same as above

src/metrics.ts
mustard-mh: Woo, it's cool that we can do it that simple way in nodejs 👍

PR #39, opened 2023-03-16
Add error report

jeanp413:
Depends on  https://github.com/gitpod-io/gitpod/pull/16871

### DIFF
--- a/src/heartbeat.ts
+++ b/src/heartbeat.ts
@@ -29,7 +29,7 @@ export class HeartbeatManager extends Disposable {
         readonly workspaceId: string,
         readonly instanceId: string,
         readonly debugWorkspace: boolean,
-        private readonly accessToken: string,
+        private readonly session: vscode.AuthenticationSession,
         private readonly publicApi: GitpodPublicApi | undefined,
         private readonly logger: vscode.LogOutputChannel,
         private readonly telemetry: TelemetryReporter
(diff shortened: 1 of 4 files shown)


src/telemetryReporter.ts
```diff
@@ -11,6 +11,10 @@ import * as vscode from 'vscode';
 const analyticsClientFactory = async (key: string): Promise<BaseTelemetryClient> => {
 	let segmentAnalyticsClient = new SegmentAnalytics(key);
 
+	const gitpodHost = vscode.workspace.getConfiguration('gitpod').get<string>('host')!;
```
mustard-mh, line 14: It should get inside function itself in case user changed gitpod host?

src/heartbeat.ts
```diff
@@ -144,7 +144,8 @@ export class HeartbeatManager extends Disposable {
             }, this.logger);
         } catch (e) {
             const suffix = wasClosed ? 'closed heartbeat' : 'heartbeat';
-            this.logger.error(`Failed to send ${suffix}, triggered by ${this.lastActivityEvent} event:`, e);
+            e.message = `Failed to send ${suffix}, triggered by event: ${this.lastActivityEvent}: ${e.message}`;
```
mustard-mh, line 147: - Wrap error message
- Use `this.logger.error(e)` to avoid error message in collected logs shows `failed to xxx: {protocol: http}`, it expect Error as its first argument

mustard-mh:
All good with locally packaged
image

Results
image
image

mustard-mh:
image

src/heartbeat.ts
mustard-mh: image

mustard-mh:
We will need better way to call `this.logger.error` -> wrap the error and make it first argument

Let's first ship this changes 👍 Thank you @jeanp413

src/telemetryReporter.ts
mustard-mh: I changed it, revert this change if it's not correct

PR #40, opened 2023-03-16
Update Nightly job to trigger it every main push

mustard-mh:
‼️ This will change nightly extension version from `YYYYMMDDHH` to `YYYYMMDDHHmm`. To unblock in hour publish

@jeanp413

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -4,12 +4,30 @@ on:
   workflow_dispatch:
     inputs:
       patchVersion:
-        description: "Format: YYYYMMDDHH"
+        description: "Format: YYYYMMDDHHmm, empty fallback to parse from current date"
+        required: false
+  push:
+    branches:
+      - main
 
 jobs:
   nightly:
     runs-on: ubuntu-latest
     steps:
+      - name: "Set patchVersion outputs"
+        id: configuration
+        shell: bash
+        run: |
+            if [[ '${{ github.event.inputs.patchVersion }}' != '' ]]; then
+              # The workflow was triggered by workflow_dispatch
+              {
+                echo "version=${{ github.event.inputs.patchVersion }}"
+              } >> $GITHUB_OUTPUT
+            else
+              {
+                echo "version=$(date +'%Y%m%d%H%M')"
+              } >> $GITHUB_OUTPUT
+            fi
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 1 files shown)


jeanp413:
Also changing to `YYYYMMDDHHmm` will make the patch version bigger than the max 32bit signed integer could cause some error in the server, [openvsx assumes 32 bit](https://github.dev/eclipse/openvsx/blob/a874a24e85653463b8cc7697d6849ea7a64f49cb/server/src/main/java/org/eclipse/openvsx/entities/ExtensionVersion.java#L252) :see_no_evil:  not sure in MS marketplace

jeanp413:
cc @akosyakov :point_up_2:

jeanp413:
> When we push to master we should have confidence in it. Any commit should work in theory. If we are not sure go via the PR? I don't think ability pushing to master means we can break it.

it works, it's just some code commented for testing that slipped

> How other extensions doing it? I'm pretty sure auto publish is not uncommon.

 it's once a day not on every commit

akosyakov:
To make the matter simpler we could publish only once a day at the time when most likely there are no changes, i.e. after JP goes to sleep :) The same way how we deploy Gitpod daily?

jeanp413:
yeah, then should we stop deploying manually and wait for auto deploy? all this time i've always deployed both at the same time

jeanp413:
@mustard-mh is there a way to check the released commit so we don't deploy the same commit each day if no changes are made :thinking:

mustard-mh:
> I wonder if there's a way to check the released commit so we don't deploy the same commit each day if no changes are made or is this fine 🤔

We can create a branch for nightly and compare the commit hash like gitlens does https://github.com/gitkraken/vscode-gitlens/blob/main/.github/workflows/cd-pre.yml#L26-L27

So for nightly, it will be once a day if we have some updates automatically

> all this time i've always deployed both at the same time after a change/PR

If we release stable changes, we should have nightly release too. release should trigger them both

akosyakov:
> We can create a branch for nightly and compare the commit hash like gitlens does https://github.com/gitkraken/vscode-gitlens/blob/main/.github/workflows/cd-pre.yml#L26-L27
> So for nightly, it will be once a day if we have some updates automatically

sounds like a plan

> If we release stable changes, we should have nightly release too. release should trigger them both

I don't think we need to couple it. Nightly means once a day as you said if there are no changes. It can be done on demand as well.

jeanp413:
Follow up PR https://github.com/gitpod-io/gitpod-vscode-desktop/pull/64

PR #42, opened 2023-03-17
Add retry dialog in onGitpodRemoteConnection

jeanp413:


mustard-mh:
@jeanp413 👍 

I'd like to remove last trigger event for sendHeartbeat in this PR, since the different event name will trigger different ErrorReporting. 

Thank you

mustard-mh:
```ts
catch (e) {
	if (e instanceof NoRunningInstanceError) {
		this.logger.error('No Running instance:', e);
		return;
	}
}
```

This one should included right? but with different dialog: workspace is not running and redirect to browser like `Restart Workspace` dis

mustard-mh:
Test restart dialog with change 
image

Results
image

image

mustard-mh:
Retry test code
image


Result
image

Assaf1978-boop:
Nice.

PR #43, opened 2023-03-21
Do some refactor and cleanup

jeanp413:


src/services/sessionService.ts
```diff
@@ -0,0 +1,184 @@
...
+        if (this.isSignedIn()) {
+            this.logger.info(`Successfully signed in`);
+        } else {
+            this.logger.error(`Failed to sign in`);
+        }
+    }
+
+    private async tryLoadSession(createIfNone: boolean) {
+        try {
+            if (this.session && !createIfNone) {
```
mustard-mh, line 124:
```suggestion
            if (this.session || !createIfNone) {
```

src/services/sessionService.ts
```diff
@@ -0,0 +1,184 @@
...
+        if (oldSession && !this.session) {
+            // sign out
+            didChange = true;
+        } else if (!oldSession && this.session) {
+            // sign in
+            didChange = true;
+        } else if (oldSession?.id !== this.session?.id) {
+            // host changed
+            didChange = true;
+        }
```
mustard-mh, line 75:
```suggestion
        // host changed, sign out, sign in
        const didChange = oldSession?.id !== this.session?.id;
```

src/services/hostService.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    async changeHost(newHost: string, force: boolean = false) {
+        if (new URL(this._gitpodHost).host !== new URL(newHost).host) {
+            const flow: UserFlowTelemetry = { flow: 'changeHost', gitpodHost: newHost };
+
+            // Don't allow to change gitpod host if we are in a remote window
+            if (!force && !!getGitpodRemoteWindowConnectionInfo(this.context)) {
+                this.notificationService.showWarningMessage(`Cannot to swith to '${newHost}' while connected to '${this._gitpodHost}'. Please close the window first`, { id: 'switch_gitpod_host_remote_window', flow });
+                return false;
+            }
```
mustard-mh, line 64:
```suggestion
    async changeHost(newHost: string, ignoreGitpodRemoteConn: boolean = true) {
        if (new URL(this._gitpodHost).host !== new URL(newHost).host) {
            const flow: UserFlowTelemetry = { flow: 'changeHost', gitpodHost: newHost };

            // Don't allow to change gitpod host if we are in a remote window
            if (ignoreGitpodRemoteConn && !!getGitpodRemoteWindowConnectionInfo(this.context)) {
                this.notificationService.showWarningMessage(`Cannot to swith to '${newHost}' while connected to '${this._gitpodHost}'. Please close the window first`, { id: 'switch_gitpod_host_remote_window', flow });
                return false;
            }
```

mustard-mh: as `force` makes me confused, I thought it was force to changeHost but actully is force to show notification

jeanp413: `ignoreGitpodRemoteConn`  being `true` by default confuses me :rofl: shouldn't it be `dontIgnoreGitpodRemoteConn` instead or `skipRemoteWindowCheck`

jeanp413: renamed to `skipRemoteWindowCheck`, does that sounds better?

src/services/sessionService.ts
jeanp413: renamed to `force` :see_no_evil: , does that sounds better?

PR #45, opened 2023-03-29
Readd `onResolveRemoteAuthority:ssh-remote` and fix for configuration default values

jeanp413:
## Description

When activating the extension early with `onResolveRemoteAuthority:ssh-remote`, configuration default values are not available yet and will return `undefined` so we hardcode the defaults

## How to test

- Ensure you don't have changed any gitpod configuration in your settings.json if so then remove it first, so default values are used
- Try connecting to a workspace in gitpod.io as this is the default value, connection should be successful and the extension is activated

### DIFF
--- a/scripts/prepare-release-build.js
+++ b/scripts/prepare-release-build.js
@@ -4,13 +4,4 @@ const fs = require("fs");
 
 const releasePackageJson = JSON.parse(fs.readFileSync('./package.json').toString());
 
-const releaseDefaultConfig = new Map([
-    ["gitpod.remote.useLocalApp", true],
-]);
-
-const gitpodConfig = releasePackageJson.contributes.configuration.find(e => e.title.toLowerCase() === 'gitpod');
-for (const [setting, value] of releaseDefaultConfig) {
-    gitpodConfig.properties[setting].default = value;
-}
-
 fs.writeFileSync('./package.release.json', JSON.stringify(releasePackageJson, undefined, '\t') + '\n');
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -1,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Gitpod. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+
+// Use these functions instead of `vscode.workspace.getConfiguration` API
+// When activating the extension early with `onResolveRemoteAuthority:ssh-remote`, default values
+// are not available yet and will return `undefined` so we hardcode the defaults here
+
+function getGitpodHost() {
+    return vscode.workspace.getConfiguration('gitpod').get<string>('host', 'https://gitpod.io/');
+}
+
+function getShowReleaseNotes() {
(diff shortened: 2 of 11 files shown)


src/remoteConnector.ts
```diff
@@ -650,9 +651,7 @@ export class RemoteConnector extends Disposable {
 		this.usePublicApi = await this.experiments.getRaw<boolean>('gitpod_experimental_publicApi', { gitpodHost: params.gitpodHost }) ?? false;
 		this.logService.info(`Going to use ${this.usePublicApi ? 'public' : 'server'} API`);
 
-		const forceUseLocalApp = getServiceURL(params.gitpodHost) === 'https://gitpod.io'
-			? (await this.experiments.get<boolean>('gitpod.remote.useLocalApp', { gitpodHost: params.gitpodHost }))!
-			: (await this.experiments.get<boolean>('gitpod.remote.useLocalApp', { gitpodHost: params.gitpodHost }, 'gitpod_remote_useLocalApp_sh'))!;
```
jeanp413, line 655: values were set to `false` for 100% users in configcat for a long time so let's remove this already to make things easier

scripts/prepare-release-build.js
```diff
@@ -4,13 +4,4 @@ const fs = require("fs");
...
 
-const releaseDefaultConfig = new Map([
-    ["gitpod.remote.useLocalApp", true],
-]);
-
-const gitpodConfig = releasePackageJson.contributes.configuration.find(e => e.title.toLowerCase() === 'gitpod');
-for (const [setting, value] of releaseDefaultConfig) {
-    gitpodConfig.properties[setting].default = value;
-}
-
```
jeanp413, line 15: related to above comment :point_up:

src/extension.ts
```diff
@@ -35,95 +35,101 @@ if (!global.fetch) {
...
+		}));
+
+		context.subscriptions.push(new ReleaseNotes(context, commandManager, logger));
+
+		// Register global commands
+		commandManager.register(new SignInCommand(sessionService));
+		commandManager.register(new ExportLogsCommand(context.logUri, notificationService, telemetryService, logger));
+
+		const remoteConnectionInfo = getGitpodRemoteWindowConnectionInfo(context);
+		telemetryService.sendTelemetryEvent('vscode_desktop_activate', {
```
akosyakov, line 103: we shoudl mode it to finally block and `success: boolean` property to indicate whether acitaved or not

src/configuration.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Gitpod. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+
+// Use these functions instead of `vscode.workspace.getConfiguration` API
```
akosyakov, line 8: Is it a problem only with default values or at user setttings are not loaded as well yet?

akosyakov (commented):
It is generally looks ok if it is indeed only default values missing not all settings, i.e. this question is most important to answer: https://github.com/gitpod-io/gitpod-vscode-desktop/pull/45#discussion_r1151554618 and test with non default settings!

PR #46, opened 2023-03-30
Add `ide_heartbeat` telemetry IDE-19

mustard-mh:
Related https://linear.app/gitpod/issue/IDE-18/align-ssh-events-for-all-clients

### DIFF
--- a/src/heartbeat.ts
+++ b/src/heartbeat.ts
@@ -12,11 +12,24 @@ import { ITelemetryService } from './services/telemetryService';
 import { SSHConnectionParams } from './remote';
 import { ISessionService } from './services/sessionService';
 import { ILogService } from './services/logService';
+import { RawTelemetryEventProperties } from './common/telemetry';
+
+const IDEHeartbeatTelemetryEvent = 'ide_heartbeat';
+interface IDEHeartbeatTelemetryData extends RawTelemetryEventProperties {
+    clientKind: 'vscode-desktop'; // 'ssh' | 'jetbrains' | 'vscode-desktop' | 'supervisor-frontend';
+    totalCount: number;
+    successfulCount: number;
+    workspaceId: string;
+    instanceId: string;
+    gitpodHost: string;
+    debugWorkspace: 'true' | 'false';
+}
 
 export class HeartbeatManager extends Disposable {
 
     static HEARTBEAT_INTERVAL = 30000;
     static EVENT_COUNTER_INTERVAL = 3600000;
+    static IDE_HEARTBEAT_INTERVAL = 900000; // 15 minutes
 
     private lastActivity = new Date().getTime();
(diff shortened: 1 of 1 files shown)


src/heartbeat.ts
```diff
@@ -12,11 +12,24 @@ import { ITelemetryService } from './services/telemetryService';
...
+
+const IDEHeartbeatTelemetryEvent = 'ide_heartbeat';
+interface IDEHeartbeatTelemetryData extends RawTelemetryEventProperties {
+    clientKind: 'vscode-desktop'; // 'ssh' | 'jetbrains' | 'vscode-desktop' | 'supervisor-frontend';
+    totalCount: number;
+    successfulCount: number;
+    workspaceId: string;
+    instanceId: string;
+    gitpodHost: string;
+    debugWorkspace: 'true' | 'false';
```
filiptronicek, line 25: Can this be a boolean to be consistent with `connectionInfo.debugWorkspace`?

src/heartbeat.ts
```diff
@@ -147,7 +180,9 @@ export class HeartbeatManager extends Disposable {
             e.message = `Failed to send ${suffix}, triggered by event: ${this.lastActivityEvent}: ${originMsg}`;
             this.logService.error(e);
             e.message = `Failed to send ${suffix}: ${originMsg}`;
-            this.telemetryService.sendTelemetryException(e, { workspaceId: this.connectionInfo.workspaceId, instanceId: this.connectionInfo.instanceId});
+            this.telemetryService.sendTelemetryException(e, { workspaceId: this.connectionInfo.workspaceId, instanceId: this.connectionInfo.instanceId });
+        } finally {
+            this.ideHeartbeatData.totalCount++;
```
akosyakov, line 185: I think both variables should be increased here in one tick.

akosyakov:
@mustard-mh Could you address the feedback and release it?

src/heartbeat.ts
mustard-mh: > Can this be a boolean to be consistent with connectionInfo.debugWorkspace?

It can be, but I just want to keep it like original field, I guess maybe it's because Mixpanel can't properly identify `undefined` and `false`

src/heartbeat.ts
akosyakov: I mean they should be increased in the same JS event loop tick, since successfulCount is increased within `await withServerApi` it does not guarantee

mustard-mh: Addressed, going to merge this PR

mustard-mh:
@iQQBot May I ask you to review this PR again? Since they might not online now

mustard-mh:
Going to test and post result here

mustard-mh:
image

PR #47, opened 2023-04-17
report telemetry to corresponding host

akosyakov:
## Description

depends on https://github.com/gitpod-io/gitpod/pull/17162

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

### DIFF
--- a/src/common/telemetry.ts
+++ b/src/common/telemetry.ts
@@ -63,12 +63,10 @@ export class BaseTelemetryAppender implements ITelemetryAppender {
 	private _exceptionQueue: Array<{ exception: Error; data: AppenderData | undefined }> = [];
 
 	// Necessary information to create a telemetry client
-	private _clientFactory: (key: string) => Promise<BaseTelemetryClient>;
-	private _key: string;
+	private _clientFactory: () => Promise<BaseTelemetryClient>;
 
-	constructor(key: string, clientFactory: (key: string) => Promise<BaseTelemetryClient>) {
+	constructor(clientFactory: () => Promise<BaseTelemetryClient>) {
 		this._clientFactory = clientFactory;
-		this._key = key;
 		if (getTelemetryLevel() !== TelemetryLevel.OFF) {
 			this.instantiateAppender();
 		}
@@ -124,20 +122,25 @@ export class BaseTelemetryAppender implements ITelemetryAppender {
 		this._exceptionQueue = [];
 	}
 
+	private isInstantiating = false;
+
 	/**
 	 * Instantiates the telemetry client to make the appender "active"
 	 */
 	instantiateAppender(): void {
-		if (this._isInstantiated) {
+		if (this.isInstantiating || this._isInstantiated) {
 			return;
 		}
+		this.isInstantiating = true;
 		// Call the client factory to get the client and then let it know it's instatntiated
-		this._clientFactory(this._key).then(client => {
+		this._clientFactory().then(client => {
 			this._telemetryClient = client;
 			this._isInstantiated = true;
 			this._flushQueues();
 		}).catch(err => {
 			console.error(err);
+		}).finally(() => {
+			this.isInstantiating = false;
 		});
 	}
 }
@@ -146,12 +149,13 @@ export class BaseTelemetryReporter extends Disposable {
 	private userOptIn = false;
 	private errorOptIn = false;
 	private _extension: vscode.Extension<any> | undefined;
+	private readonly appenders = new Map<string, ITelemetryAppender>();
 
 	constructor(
 		private extensionId: string,
 		private extensionVersion: string,
-		private telemetryAppender: ITelemetryAppender,
 		private osShim: { release: string; platform: string; architecture: string },
+		private readonly clientFactory: (gitpodHost: string) => Promise<BaseTelemetryClient>
 	) {
 		super();
 
@@ -173,7 +177,7 @@ export class BaseTelemetryReporter extends Disposable {
 		this.userOptIn = telemetryLevel === TelemetryLevel.ON;
 		this.errorOptIn = telemetryLevel === TelemetryLevel.ERROR || this.userOptIn;
 		if (this.userOptIn || this.errorOptIn) {
-			this.telemetryAppender.instantiateAppender();
+			this.appenders.forEach(appender => appender.instantiateAppender());
 		}
 	}
 
@@ -354,11 +358,11 @@ export class BaseTelemetryReporter extends Disposable {
 	 * @param eventName The name of the event
 	 * @param properties The properties to send with the event
 	 */
-	public sendTelemetryEvent(eventName: string, properties?: TelemetryEventProperties): void {
+	public sendTelemetryEvent(gitpodHost: string, eventName: string, properties?: TelemetryEventProperties): void {
 		if (this.userOptIn && eventName !== '') {
 			properties = { ...properties, ...this.getCommonProperties() };
 			const cleanProperties = this.cloneAndChange(properties, (_key: string, prop: string) => this.anonymizeFilePaths(prop, false));
-			this.telemetryAppender.logEvent(`${eventName}`, { properties: this.removePropertiesWithPossibleUserInfo(cleanProperties) });
+			this.getAppender(gitpodHost).logEvent(`${eventName}`, { properties: this.removePropertiesWithPossibleUserInfo(cleanProperties) });
 		}
 	}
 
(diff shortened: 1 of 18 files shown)


src/common/telemetry.ts
```diff
@@ -146,35 +144,15 @@ export class BaseTelemetryReporter extends Disposable {
...
-	/**
-	 * Updates whether the user has opted in to having telemetry collected
-	 */
-	private updateUserOptStatus(): void {
-		const telemetryLevel = getTelemetryLevel();
-		this.userOptIn = telemetryLevel === TelemetryLevel.ON;
-		this.errorOptIn = telemetryLevel === TelemetryLevel.ERROR || this.userOptIn;
-		if (this.userOptIn || this.errorOptIn) {
-			this.telemetryAppender.instantiateAppender();
-		}
```
jeanp413, line 177: why remove this? we should respect user choice, i always have telemetry disabled for example

src/common/telemetry.ts
```diff
@@ -146,35 +144,15 @@ export class BaseTelemetryReporter extends Disposable {
 	private userOptIn = false;
 	private errorOptIn = false;
 	private _extension: vscode.Extension<any> | undefined;
+	private readonly appenders = new Map<string, ITelemetryAppender>();
```
jeanp413, line 147: I saw a thread in slack about supporting multiple host, but that should be discussed and addressed in a separate PR for the whole extension, not mix changes here

src/services/telemetryService.ts
```diff
@@ -98,26 +104,25 @@ export interface UserFlowTelemetry extends TelemetryOptions {
 }
 
 export interface ITelemetryService {
-	sendTelemetryEvent(eventName: string, properties?: TelemetryEventProperties): void;
-	sendRawTelemetryEvent(eventName: string, properties?: RawTelemetryEventProperties): void;
-	sendTelemetryException(error: Error, properties?: TelemetryEventProperties): void;
+	sendTelemetryEvent(gitpodHost: string, eventName: string, properties?: TelemetryEventProperties): void;
+	sendRawTelemetryEvent(gitpodHost: string, eventName: string, properties?: RawTelemetryEventProperties): void;
+	sendTelemetryException(gitpodHost: string, error: Error, properties?: TelemetryEventProperties): void;
```
jeanp413, line 109: would it be possible to add gitpodHost automatically in the `${gitpodHost}/analytics` endpoint as the body is just json instead of doing it in every call? for development it doesn't matter as we know which host we are testing manually

src/extension.ts
```diff
@@ -59,7 +61,12 @@ export async function activate(context: vscode.ExtensionContext) {
 
 		logger.info(`${extensionId}/${packageJSON.version} (${os.release()} ${os.platform()} ${os.arch()}) vscode/${vscode.version} (${vscode.env.appName})`);
 
-		telemetryService = new TelemetryService(extensionId, packageJSON.version, packageJSON.segmentKey, logger!);
+		let segmentKey = packageJSON.segmentKey;
+		if (context.extensionMode !== vscode.ExtensionMode.Production) {
+			// in production mode we use the Gitpod server key
+			segmentKey = "";
```
jeanp413, line 67: would it be better to always send the segment key? with the key we can at least do a simple validation in the analytics endpoint  so it's not trivial no exploit (even when the key is easy to get)

src/services/telemetryService.ts
```diff
@@ -82,7 +88,7 @@ const analyticsClientFactory = async (key: string, logger: vscode.LogOutputChann
 };
 
 interface TelemetryOptions {
-	gitpodHost?: string;
+	gitpodHost: string;
```
jeanp413, line 91: wouldn't it be possible to add `gitpodHost` automatically  in the `${gitpodHost}/analytics`  endpoint as the body is just json, this will work for all events not just the ones required explicitly by UserFlowTelemetry and we wouldn't have to worry about forgetting sending it

src/common/telemetry.ts
akosyakov: it is not about multi host, it is about analytics to corresponding host

src/extension.ts
akosyakov: - we cannot trust keys anyway bundled in extensions, anyone can reverse engineer it, we are going to add rate limitting in case of abuse on endpoint itself
- for trusted events from Gitpod's component we use key indeed

src/services/telemetryService.ts
akosyakov: I did not get it, sorry. Could you elaborate what changes are necessary?

src/extension.ts
akosyakov: actually it does not work, Node.js client needs some key, [internal discussion](https://gitpod.slack.com/archives/C01KPEPKX0A/p1681897024307559)

src/services/telemetryService.ts
jeanp413: [Here](https://github.com/gitpod-io/gitpod/blob/d5525e3dc938428af34c2dd187986e1fb4c35d8f/components/proxy/plugins/analytics/analytics.go#L72-L90) intercepet the request, parse the json body and add `gitpodHost` property instead of doing it for every event in each client

src/common/telemetry.ts
jeanp413: You can just use `getGitpodHost` which reads it from setting, no need to pass it around as the first thing we do  is update the settings when trying to connect to a new workspace before sending any telemetry

> it is reporting about analytics to corresponding host, never around or to another.

We only support one host currently, the first action when connecting to a new workspace is [sign in](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/ff65e7992ff40472ac4c7e24a4be05db1f664aad/src/remoteConnector.ts#L641) which internally updates `gitpodHost` setting, also there's a [notification when user has a remote workspace already connected](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/ff65e7992ff40472ac4c7e24a4be05db1f664aad/src/remoteSession.ts#L44-L50) which can be improved

src/services/telemetryService.ts
akosyakov: Ah, for property, we should completely remove it. It is not necessary anymore.

src/common/telemetry.ts
akosyakov: > You can just use getGitpodHost

I don't think it is true. For instance in https://github.com/gitpod-io/gitpod-vscode-desktop/blob/ff65e7992ff40472ac4c7e24a4be05db1f664aad/src/authentication/authentication.ts#L218 gitpod service should be fixed during all execution, but since code is async if for some reason config is changed during event ticks telemetry will leak to another host, we should make sure that never happens.

src/services/telemetryService.ts
```diff
@@ -3,15 +3,38 @@
...
+		settings.path = '/analytics' + settings.path;
+	} else {
+		if (serviceUrl.host !== "gitpod.io" && !serviceUrl.host.endsWith(".gitpod-dev.com")) {
+			logger.warn(`No telemetry: dedicated installations should send data always to own endpoints, host: ${serviceUrl.host}`);
+			return {
+				logEvent: () => { },
+				logException: () => { },
+				flush: () => { },
+			}
+		}
```
jeanp413, line 31: So if we are running this extension from sources and connect to dedicated, we cannot send telemetry to ide playground for testing and need to switch to gitpod.io :thinking: why is that? I think it should be fine to do so

src/services/telemetryService.ts
```diff
@@ -102,26 +117,26 @@ export interface UserFlowTelemetry extends TelemetryOptions {
...
-		});
+		}, gitpodHost => analyticsClientFactory(gitpodHost, segmentKey, logger));
 	}
 
 	sendUserFlowStatus(status: string, flow: UserFlowTelemetry): void {
-		const properties: TelemetryOptions = { ...flow, status };
+		const properties: Partial<TelemetryOptions> = { ...flow, status };
 		delete properties['flow'];
-		this.sendRawTelemetryEvent('vscode_desktop_' + flow.flow, properties);
+		delete properties['gitpodHost'];
```
jeanp413, line 139: Not sure about this, we still have self hosted installation events and I use this to monitor, how are we gonna differentiate

src/services/telemetryService.ts
mustard-mh: @akosyakov  Yep, `else` should never happen.

I'm not sure about the target:

🅰️ 
1. Source -> Local trace
2. SaaS -> SaaS
3. dev -> dev
4. Others -> their gitpodHost

or 🅱️ 
1. Source -> Local trace
2. SaaS -> SaaS
3. dev -> dev
4. Others -> omit

==============
Right now, we want and are going with 🅰️ . So we can remove `else`?

akosyakov: Telemetry from dedicated should always be reported to a corresponding dedicated cell. We are legally not allowed to distribute it anywhere else.

mustard-mh: So else is working for the case we build from source and use dedicated? ok then

PR #48, opened 2023-04-25
Stop send `heartbeat_delta` if workspace is not running

mustard-mh:
https://linear.app/gitpod/issue/IDE-43/dont-send-heartbeat-delta-telemetry-after-workspace-is-stopped

### DIFF
--- a/src/heartbeat.ts
+++ b/src/heartbeat.ts
@@ -139,6 +139,8 @@ export class HeartbeatManager extends Disposable {
                 } else {
                     this.logService.trace('Stopping heartbeat as workspace is not running');
                     this.stopHeartbeat();
+                    this.stopEventCounter();
+                    this.sendEventData();
                 }
             }, this.logService);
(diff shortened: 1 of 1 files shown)


src/heartbeat.ts
```diff
@@ -139,6 +139,8 @@ export class HeartbeatManager extends Disposable {
                 } else {
                     this.logService.trace('Stopping heartbeat as workspace is not running');
                     this.stopHeartbeat();
+                    this.stopEventCounter();
```
akosyakov, line 142: When we stop, we should report the final analytics event which reports the final delta.

I think we should also merge `vscode_desktop_heartbeat_delta` into `ide_heartbeat` by adding more context in case of VS Code Desktop.

mustard-mh:
Close since we move stop and delta telemetry to PR https://github.com/gitpod-io/gitpod-vscode-desktop/pull/46

PR #49, opened 2023-04-26
Graceful restart local ssh daemon

mustard-mh:
- [ ] It should rebase after PR https://github.com/gitpod-io/gitpod-vscode-desktop/pull/50 merged

This PR changed 

#### 1️⃣  add restart logic in daemon
- If there's no connection to local-ssh
- Get extension version from extension ipc
- Compare with daemon running extension version
- If it's greater than current one, ask extension ipc to restart daemon later (trigger `onDidDaemonRestartRequired` event)
- Exit with proper exit code

#### 2️⃣  where we start daemon
 from
- Extension active
- Extension ipc service active and every time we are trying to ping local-ssh ipc

To 
- Extension active
- Extension ipc service received `onDidDaemonRestartRequired` event
- Trying to start a remote connection

## Related Issue(s)

Fixes https://linear.app/gitpod/issue/IDE-54

### DIFF
--- a/src/local-ssh/ipc/extension.ts
+++ b/src/local-ssh/ipc/extension.ts
@@ -68,7 +68,7 @@ export class ExtensionServiceImpl implements ExtensionServiceImplementation {
             }
             const userId = this.sessionService.getUserId();
             const workspaceId = request.workspaceId;
-            
+
             const gitpodHost = this.hostService.gitpodHost;
             const usePublicApi = await this.experiments.getUsePublicAPI(gitpodHost);
             const [workspace, ownerToken] = await withServerApi(accessToken, gitpodHost, svc => Promise.all([
@@ -149,7 +149,7 @@ export class ExtensionServiceImpl implements ExtensionServiceImplementation {
     async getCurrentExtensionVersion(_request: getCurrentExtensionVersionRequest, _context: CallContext): Promise<{ version?: string | undefined; }> {
         return { version: this.extensionVersion };
     }
-    
+
     async tryRestartDaemon(_request: TryRestartDaemonRequest, _context: CallContext): Promise<{}> {
         this.logService.info('daemon requested to restart');
         this._onDidDaemonRestartRequired.fire();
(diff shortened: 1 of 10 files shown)


mustard-mh:
@jeanp413 I updated PR description, not test it yet, will continue work on it tomorrow, but you can review it first if you have time since it's almost complete 🙏

jeanp413:
Let's discuss a little bit more:
- what's the delay of starting the daemon on demand?
- right now I only see two cases:
  - on activate **if restoring a remote window** this means user closes vscode completely so daemon is not running
  - on uriHandler when trying to connect to a new workspace, this will be the first remote window, we "save" here time if user has already open vscode but hown much? other case is vscode was completely closed and is started by the uri handler and we need to start daemon at the same time

at any other time daemon don't need to be running I think, I'm I missing some other case

src/local-ssh/ipc/extension.ts
```diff
@@ -185,22 +214,26 @@ export class ExtensionServiceServer extends Disposable {
...
-            this.backoffActive();
+            this.setExtensionActivedForDaemon();
         }).catch(e => {
             this.logService.error(e, `extension ipc service server failed to listen with id: ${this.id}`);
         });
     }
 
-    private async backoffActive() {
+    private async setExtensionActivedForDaemon() {
+        while (true) {
```
akosyakov, line 224: should not we cancel such things that our extension don't keep doing anything if deactivated

src/local-ssh/server.ts
```diff
@@ -59,6 +67,28 @@ export class LocalSSHGatewayServer {
 		return session;
 	}
 
+	private async tryReloadDaemon() {
+		if (this.clientCount > 0) {
```
akosyakov, line 71: Does it mean we would run several daemons in parallel? i.e. if some window already upgraded and assumes a never version of daemon, but others are not yet?

mustard-mh: It's connection count (need to rename it)

PR #50, opened 2023-04-26
Close conn based on piped conn IDE-50

mustard-mh:
## Description

Same with title

## Related Issue(s)

Related https://linear.app/gitpod/issue/IDE-50

## How to test

- Install extension locally
- Connect to any workspaces
- Exec `gp stop` to stop workspace
- It should close connection to local-ssh daemon

|SSH Gateway| Local SSH |
|:-:|:-:|
|SCR-20230426-vi0|SCR-20230426-w5l|

### DIFF
--- a/src/daemonStarter.ts
+++ b/src/daemonStarter.ts
@@ -61,7 +61,7 @@ export async function tryStartDaemon(logService: ILogService, options?: Partial<
     const daemon = spawn(process.execPath, args, {
         detached: true,
         stdio: 'ignore',
-        env: process.env
+        env: { ...process.env, ELECTRON_RUN_AS_NODE: '1' },
     });
     daemon.unref();
     return daemon;
--- a/src/local-ssh/ipc/localssh.ts
+++ b/src/local-ssh/ipc/localssh.ts
@@ -81,10 +81,10 @@ export class LocalSSHServiceImpl implements LocalSSHServiceImplementation {
     }
 
     public async getWorkspaceAuthInfo(workspaceId: string) {
-        return retryWithStop(async (stop) => {
-            const getAuthInfo = async (id: string, client: Client<ExtensionServiceDefinition>) => {
+        return retryWithStop((stop) => {
+            const getAuthInfo = (id: string, client: Client<ExtensionServiceDefinition>) => {
                 try {
-                    return await client.getWorkspaceAuthInfo({ workspaceId });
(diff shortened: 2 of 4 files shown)


jeanp413 (approved):

src/local-ssh/ipc/localssh.ts
```diff
@@ -81,10 +81,10 @@ export class LocalSSHServiceImpl implements LocalSSHServiceImplementation {
     }
 
     public async getWorkspaceAuthInfo(workspaceId: string) {
-        return retryWithStop(async (stop) => {
-            const getAuthInfo = async (id: string, client: Client<ExtensionServiceDefinition>) => {
+        return retryWithStop((stop) => {
+            const getAuthInfo = (id: string, client: Client<ExtensionServiceDefinition>) => {
                 try {
-                    return await client.getWorkspaceAuthInfo({ workspaceId });
+                    return client.getWorkspaceAuthInfo({ workspaceId });
```
mustard-mh, line 87: If we return the async function, try catch will not work, I will revert this change

src/daemonStarter.ts
```diff
@@ -61,8 +61,8 @@ export async function tryStartDaemon(logService: ILogService, options?: Partial<
     const daemon = spawn(process.execPath, args, {
         detached: true,
         stdio: 'ignore',
-        env: process.env
+        env: { ...process.env, ELECTRON_RUN_AS_NODE: '1' },
```
mustard-mh, line 64: `ELECTRON_RUN_AS_NODE` what will this affect? @jeanp413

src/local-ssh/server.ts
```diff
@@ -72,19 +72,17 @@ export class LocalSSHGatewayServer {
 				let pipeSession: SshClientSession;
 				this.clientCount += 1;
 				session.onAuthenticating((e) => {
-					e.authenticationPromise = new Promise((resolve, reject) => {
-						this.authenticateClient(e.username!).then(async s => {
+					e.authenticationPromise = this.authenticateClient(e.username!).then(s => {
```
mustard-mh, line 75: Code looks weird, I'll test it @jeanp413

src/daemonStarter.ts
jeanp413: Looks the same in the end as I saw `--enable-electron-run-as-node` was used automatically too, but I saw this used in vscode extensions so better to be explicit

PR #51, opened 2023-04-27
Only show workspace stopping notification if window is correct remote window

iQQBot:
## Description

Only show workspace stopping notification if window is correct remote window

This will fixes IDE-58

## Related Issue(s)

Fixes #

## How to test

1. open workspace in vscode desktop
2. stop workspace
3. try to reconnect 
4. the notification `workspace is not running, current phase ...` should always show in correct remote window.


mustard-mh (approved):
Nice 👍 

|Before|After|
|:-:|:-:|
|SCR-20230428-1al|SCR-20230428-18i|

PR #52, opened 2023-04-27
Try to kill zombie daemon IDE-59

mustard-mh:
Linear https://linear.app/gitpod/issue/IDE-59

It will force kill daemon version where is less than `0.0.3` since it will cause zombie daemon

## How to Test

**Platforms**
- [x] Windows ✅ stable
- [x] Linux ✅ stable
- [x] macOS ✅  stable + insiders

**Test Steps**

- Package extension to `xxx.visx`
- Install extension v0.0.100, close window and start one in 10 seconds
- Quit VSCode Desktop, check if daemon process is still running after 10 seconds `watch -n 1 "ps aux | grep '[d]aemon.js'"`
- Open VSCode Desktop, and install packaged extension locally
- Reload window, previours version should force exit
- Try again, with packaged extension
- It should not force kill daemon


mustard-mh:
Tested in Windows Linux macOS

mustard-mh:
Addressed

I'm going to merge this PR

jeanp413:
We will revert this code after a few days right? Feels a little bit complex which the version check and all, maybe just force kill in all cases in the end gitpod team is using it

PR #53, opened 2023-04-27
Remove deprecated failureReason field and add error report

iQQBot:
## Description

Remove the deprecated `failureReason` field

## Related Issue(s)

Fixes #

## How to test

1. start a workspace
2. trying to open this workspace in desktop
3. during this time quickly do `rm -rf ~/.ssh/authorized_keys` a few times, you can use `while`, it will cause tunnel ssh auth failed.
4. check `ide-metrics` logs


jeanp413:
This is sending the whole stack as part of the error message and it's not anonymizing file paths, we need to fix that

PR #57, opened 2023-05-10
Use ssh proxy command instead of daemon IDE-72

mustard-mh:
## Related Issue(s)

Fixes https://linear.app/gitpod/issue/IDE-72

## How to Test

- Pack files into one `yarn vscode:prepublish`
- Change `.vscode/launch.json` file to remove `"preLaunchTask": "npm: watch"`
- Run code with VSCode -> Run and Debug -> Extension
- It should able to connect to workspaces

You can try it via dedicated Gitpod since we have enabled it all

- [x] macOS
- [x] Windows
- [x] Linux

### DIFF
--- a/src/local-ssh/proxy.ts
+++ b/src/local-ssh/proxy.ts
@@ -245,7 +245,7 @@ class WebSocketSSHProxy {
                 }
                 throw new FailedToGetAuthInfoError();
             });
-        }, 200, 10);
+        }, 200, 50);
     }
 
     async establishTunnel(workspaceInfo: GetWorkspaceAuthInfoResponse) {


src/local-ssh/common.ts
```diff
@@ -16,71 +14,8 @@ export function getHostKeyFingerprint(): Buffer {
...
-				} else {
-					resolve(true);
-				}
-			});
-		}),
-		new Promise<boolean>(resolve => setTimeout(() => resolve(false), timeout))
-	]);
-}
-
 export const GitpodDefaultLocalhost = 'lssh.gitpod.io';
```
jeanp413, line 17: Rename to `vss.gitpod.io` and `vsi.gitpod.io`

src/local-ssh/starter.sh
```diff
@@ -0,0 +1 @@
+ELECTRON_RUN_AS_NODE=1 "$@"
```
jeanp413, line 1: Do we need `#!/usr/bin/env bash` at the top? vscode scripts have them

src/local-ssh/starter.bat
```diff
@@ -0,0 +1,5 @@
+@ECHO OFF
```
jeanp413, line 1: vscode script also have `setlocal`, should we add it?

src/local-ssh/ipc/extension.ts
```diff
@@ -167,92 +155,20 @@ export class ExtensionServiceServer extends Disposable {
...
+        this.logService.debug('going to try active extension ipc service server on port ' + port);
+        this.server.listen('127.0.0.1:' + port).then(() => {
+            this.logService.info('extension ipc service server started to listen');
         }).catch(e => {
-            this.logService.error(e, `extension ipc service server failed to listen with id: ${this.id}`);
+            this.logService.debug(`extension ipc service server failed to listen`, e);
+            // TODO(lssh): listen to port and wait until disconnect and try again
+            timeout(1000).then(() => {
+                this.tryActive();
+            });
```
jeanp413, line 167: Maybe we need to connect to ensure the port is used by ourselves and not another app, if it's another app then show a message to user

src/local-ssh/logger.ts
jeanp413, line 1: As discussed, no need for it anymore, let's just report the corresponding error

src/ssh/sshConfig.ts
```diff
@@ -73,6 +74,46 @@ export default class SSHConfiguration {
...
     }
 
+    static async includeLocalSSHConfig(scopeName: string, configContent: string): Promise<boolean> {
+        const render = Handlebars.compile(`## START GITPOD {{scopeName}}
+### This section is managed by Gitpod. Any manual changes will be lost.
+
+{{{configContent}}}
+
+## END GITPOD {{scopeName}}`);
+        const newContent = render({ scopeName, configContent });
```
jeanp413, line 84: No need `Handlebars`, template strings  ``` `${value}` ``` is enough

src/ssh/sshConfig.ts
```diff
@@ -73,6 +74,46 @@ export default class SSHConfiguration {
...
+            } catch (err) {
+                // ignore
+                return false;
+            }
+        };
+        const ok = await findAndReplaceScope(getSSHConfigPath());
+        if (ok) {
+            return true;
+        }
+        return await findAndReplaceScope(systemSSHConfig);
```
jeanp413, line 114: We should only update user config, if the file don't exist then let's try creating it, if fails show error to user

src/ssh/sshConfig.ts
jeanp413: Also we should create a `gitpod` folder and put our config there, in the main config we just add `Include "gitpod/config"`

src/services/localSSHService.ts
```diff
@@ -0,0 +1,83 @@
...
+        const newExtIpcPort = Configuration.getLocalSshExtensionIpcPort();
+        const appName = vscode.env.appName.includes('Insiders') ? 'insiders' : 'stable';
+        const starter = process.platform === 'win32' ? locations.bat : locations.sh;
+        const configContent = await this.getLocalSSHConfig(appName, [this.hostService.gitpodHost], starter, locations.js, newExtIpcPort);
+        this.isSupportLocalSSH = await SSHConfiguration.includeLocalSSHConfig(appName, configContent);
+    }
+
+    private async getLocalSSHConfig(scopeName: string, hosts: string[], starter: string, jsLocation: string, extIpcPort: number) {
+        hosts = hosts.map(host => host.replace(/^[^:]+:\/\//, ''));
+        const render = Handlebars.compile(`{{#each hosts}}
```
jeanp413, line 56: Same, no need for Handlebars

src/local-ssh/starter.sh
filiptronicek: I think so 😄 

image

src/local-ssh/sshTunnel.ts
```diff
@@ -45,7 +45,7 @@ export class SupervisorSSHTunnel {
 	constructor(
 		private readonly logger: ILogService,
 		readonly workspaceInfo: GetWorkspaceAuthInfoResponse,
-		private readonly localsshService: LocalSSHServiceImpl,
+		private extensionIpc: Client<ExtensionServiceDefinition>,
 	) { }
 
 	private createPrivateKey(): Promise<string> {
```
jeanp413, line 51: Let's move this to the extension side, so the client ask for owner token and sshkey at the same time

src/ssh/sshConfig.ts
mustard-mh: I tried Include, but it's not working https://github.com/gitpod-io/gitpod-vscode-desktop/pull/57/commits/1f143b84a1c677be80156f179d31aff91cecb204

src/local-ssh/sshTunnel.ts
mustard-mh: I don't understand, we use extensionIpc here to report error only

sshkey is what we ask supervisor to create, we also need to tunnel supervisor here. I'm not going to change it

jeanp413: This github comment UI is not clear :facepalm: , I'm referring to `createPrivateKey`

mustard-mh: Yes, I mean they are doing the same functionality - tunnel ssh, extension don't need to know how we can tunnel to supervisor.

src/local-ssh/starter.sh
mustard-mh: Good catch 👍 , I checked vscode repo, they are using `#!/usr/bin/env bash` I will switch to use it to align our changes (same with .bat file `setlocal`)

src/ssh/sshConfig.ts
jeanp413: I added it seems to be working on my machine :crossed_fingers:

src/ssh/sshConfig.ts
```diff
@@ -73,6 +74,84 @@ export default class SSHConfiguration {
...
+        let content = '';
+        if (await exists(configPath)) {
+            try {
+                content = (await fs.promises.readFile(configPath, 'utf8')).trim();
+            } catch (e) {
+                throw new Error(`Could not read ssh config file at ${configPath}: ${e}`);
+            }
+        }
+
+        const scopeRegex = new RegExp(`START GITPOD INTEGRATION.+Include "gitpod/config".+END GITPOD INTEGRATION`, 's');
```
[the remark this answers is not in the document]
jeanp413: We can take a look tomorrow, so you can write to `~ssh/config` but not create "~ssh/gitpod/config"?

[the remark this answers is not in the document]
jeanp413: Works on my mac too

src/ssh/sshConfig.ts
```diff
@@ -73,6 +74,84 @@ export default class SSHConfiguration {
         return new SSHConfiguration(config);
     }
 
+    static async loadGitpodSSHConfig(): Promise<SSHConfiguration> {
```
akosyakov, line 77: nothing to change just to my curiosity what is advantage/disadvantage of using vscode.fs api instead? VS Code should be better aware of changes then?

src/services/localSSHService.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    private getHostSSHConfig(host: string, launcher: string, proxyScript: string, extIpcPort: number) {
+        return {
+            Host: '*.' + getLocalSSHDomain(host),
+            StrictHostKeyChecking: 'no',
+            ProxyCommand: `"${launcher}" "${process.execPath}" "${proxyScript}" --ms-enable-electron-run-as-node %h ${extIpcPort} "${Configuration.getLocalSSHLogPath()}"`
+        };
+    }
+
+    private async copyProxyScript() {
```
akosyakov, line 84: again just to my information:
- why do we need to copy cannot we execute from extensions dir?
- how upgrade is happening? we copy on each window?

jeanp413:
@mustard-mh why we need to update telemetry code? also we will removing that code soon see https://linear.app/gitpod/issue/IDE-76

src/services/localSSHService.ts
jeanp413: - extension dir is not stable it changes with every update as version is part of the folder name
- yep, maybe we can have a control file to keep track we latest version, but not sure if it's worth it, we need to check if all files exist either way to ensure no edge cases of somehow one of them missing for any reason, I'm also trying to reduce bundle size of proxy script so copy is fast right now is 1.4MB :grimacing:

src/ssh/sshConfig.ts
jeanp413: not sure, maybe it has some advantage when reading/writing on workspace files but for files outside workspace I image is the same, also it will work for other special vscode schemas like remote but we don't need it here

src/local-ssh/sshTunnel.ts
jeanp413: > extension don't need to know how we can tunnel to supervisor or even know that we are using tunnel or ssh gateway 

not sure I'm following it will just call the supervisor api, so proxy.js get all required credentials to authenticate and not do any side request that has to be handled and then reported back to extension

jeanp413:
@mustard-mh I tested and it's working for all OSes, one last issue is that the node process launched from the `proxylauncher` script is leaking when the ssh connection is closed, in linux I could fix it by using `exec` but not in windows, maybe the proper way to fix it is that the `proxy` script just end execution when the stdin/stdout stream is broken but right now that's not the case maybe a bug somewhere in the ssh library :thinking:

mustard-mh:
> one last issue is that the node process launched from the proxylauncher script is leaking when the ssh connection is closed

Do we need to worry about it? Users can get the script in theirs ssh settings anyway.

src/services/localSSHService.ts
```diff
@@ -77,7 +77,7 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
         return {
             Host: '*.' + getLocalSSHDomain(host),
             StrictHostKeyChecking: 'no',
-            ProxyCommand: `"${launcher}" "${process.execPath}" "${proxyScript}" --ms-enable-electron-run-as-node %h ${extIpcPort} "${Configuration.getLocalSSHLogPath()}"`
+            ProxyCommand: `"${launcher}" "${process.execPath}" "${proxyScript}" --ms-enable-electron-run-as-node %h ${extIpcPort} %r`
```
mustard-mh, line 80: username is not needed, we can get it from `onAuthenticating`, I will revert this part

jeanp413:
> Do we need to worry about it? Users can get the script in theirs ssh settings anyway.

I meant the `proxy.js` process keeps running after you close vscode window

jeanp413:
Pushed a workaround, could you give a last check that when you close the vscode window the `proxy.js` process dies too and then merge and release, we can keep fixing on master branch over the week

mustard-mh:
> I meant the proxy.js process keeps running after you close vscode window

It's because logic to exit processes is [deleted](https://github.com/gitpod-io/gitpod-vscode-desktop/pull/57/commits/c2d6f258c144489b5ec5ed00cc9e1734e3982de6#diff-40d5c6acef496411e05e19f7a7abdac09956e61951a63f87b1751bb0ae1ed443R118-R120), will revert and add it back

```
session.onClosed(() => {
       process.exit(0);
});
```

mustard-mh:
Tested, going to merge and release

akosyakov:
@mustard-mh let everybody know when it is good to try in VS Code Desktop (Stable and Insiders) 🙏

PR #61, opened 2023-05-15
Dynamic experiments host

filiptronicek:
Make the URL host of the experiments client configurable with the good ol' `gitpod.host`.

Fixes [IDE-75](https://linear.app/gitpod/issue/IDE-75/[code-desktop]-align-usage-of-configcat-for-dedicated)

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
 	},
 	"main": "./out/extension.js",
 	"segmentKey": "YErmvd89wPsrCuGcVnF2XAl846W9WIGl",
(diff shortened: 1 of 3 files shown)


akosyakov:
@filiptronicek Does it guarantee that in all cases we will use the proper host? Are there cases there gitpod host is only changed after feature is resolved? if so in such cases it will use a wrong one.

filiptronicek:
@akosyakov I think this shouldn't be a problem, because we resolve experiments just in time they're needed (e.g. we resolve `getUsePublicAPI` and `getUseLocalSSHServer` upon `handleUri`). Because if you change the `gitpod.host`, the experiments client is updated right away it should always be correct.

PR #62, opened 2023-05-15
Remove redundant `activationEvents`

filiptronicek:
Fix this warning:
```
This activation event can be removed for extensions targeting engine version ^1.75 as VS Code will generate these automatically from your package.json contribution declarations.
```

The message comes from [here](https://github.com/microsoft/vscode/blob/b7c81ac7ccf1f94766a1bf1d1e0e4b4cfac5db66/extensions/extension-editing/src/extensionLinter.ts#L38)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,12 +34,7 @@
 	},
 	"activationEvents": [
 		"onResolveRemoteAuthority:ssh-remote",
-		"onCommand:gitpod.syncProvider.add",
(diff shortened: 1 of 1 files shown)


filiptronicek:
> but we target only 1.74 so it is not applicable to us?

my bad 🤦, I'll close this and we can reopen once we bump up the engine

PR #63, opened 2023-05-16
Proper analytics and error reporting IDE-77

akosyakov:
## Description

Screenshot 2023-05-16 at 17 39 32

## Related Issue(s)

Fixes #

### DIFF
--- a/src/local-ssh/ipc/extensionServiceServer.ts
+++ b/src/local-ssh/ipc/extensionServiceServer.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ExtensionServiceDefinition, ExtensionServiceImplementation, GetWorkspaceAuthInfoRequest, GetWorkspaceAuthInfoResponse, SendErrorReportRequest, SendLocalSSHUserFlowStatusRequest, SendLocalSSHUserFlowStatusRequest_Code, SendLocalSSHUserFlowStatusRequest_ConnType, SendLocalSSHUserFlowStatusRequest_Status } from '../../proto/typescript/ipc/v1/ipc';
+import { ExtensionServiceDefinition, ExtensionServiceImplementation, GetWorkspaceAuthInfoRequest, GetWorkspaceAuthInfoResponse, SendErrorReportRequest, SendLocalSSHUserFlowStatusRequest } from '../../proto/typescript/ipc/v1/ipc';
 import { Disposable } from '../../common/dispose';
 export { ExtensionServiceDefinition } from '../../proto/typescript/ipc/v1/ipc';
 import { withServerApi } from '../../internalApi';
@@ -49,28 +49,28 @@ class ExtensionServiceImpl implements ExtensionServiceImplementation {
 
     }
 
-	private async getWorkspaceSSHKey(ownerToken:string,workspaceId: string, workspaceHost: string) {
-		const workspaceUrl = `https://${workspaceId}.${workspaceHost}`;
-		const metadata = new BrowserHeaders();
-		metadata.append('x-gitpod-owner-token', ownerToken);
-		const client = new ControlServiceClient(`${workspaceUrl}/_supervisor/v1`, { transport: NodeHttpTransport() });
+    private async getWorkspaceSSHKey(ownerToken: string, workspaceId: string, workspaceHost: string) {
+        const workspaceUrl = `https://${workspaceId}.${workspaceHost}`;
+        const metadata = new BrowserHeaders();
+        metadata.append('x-gitpod-owner-token', ownerToken);
+        const client = new ControlServiceClient(`${workspaceUrl}/_supervisor/v1`, { transport: NodeHttpTransport() });
 
-		const privateKey = await new Promise<string>((resolve, reject) => {
-			client.createSSHKeyPair(new CreateSSHKeyPairRequest(), metadata, (err, resp) => {
-				if (err) {
-					return reject(err);
-				}
-				resolve(resp!.toObject().privateKey);
-			});
-		});
+        const privateKey = await new Promise<string>((resolve, reject) => {
+            client.createSSHKeyPair(new CreateSSHKeyPairRequest(), metadata, (err, resp) => {
+                if (err) {
+                    return reject(err);
+                }
+                resolve(resp!.toObject().privateKey);
+            });
+        });
 
-		const parsedResult = ssh2.utils.parseKey(privateKey);
-		if (parsedResult instanceof Error || !parsedResult) {
-			throw new Error('Error while parsing workspace SSH private key');
-		}
+        const parsedResult = ssh2.utils.parseKey(privateKey);
+        if (parsedResult instanceof Error || !parsedResult) {
+            throw new Error('Error while parsing workspace SSH private key');
+        }
 
-		return (parsedResult as ParsedKey).getPrivatePEM();
-	}
+        return (parsedResult as ParsedKey).getPrivatePEM();
+    }
 
     async getWorkspaceAuthInfo(request: GetWorkspaceAuthInfoRequest, _context: CallContext): Promise<GetWorkspaceAuthInfoResponse> {
         try {
@@ -102,7 +102,7 @@ class ExtensionServiceImpl implements ExtensionServiceImplementation {
             const workspaceHost = url.host.substring(url.host.indexOf('.') + 1);
             const instanceId = (usePublicApi ? (workspace as Workspace).status?.instance?.instanceId : (workspace as WorkspaceInfo).latestInstance?.id) as string;
 
-            const sshkey = await this.getWorkspaceSSHKey(ownerToken,workspaceId,workspaceHost);
+            const sshkey = await this.getWorkspaceSSHKey(ownerToken, workspaceId, workspaceHost);
 
             return {
                 gitpodHost,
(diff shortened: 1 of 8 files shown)


src/local-ssh/proxy.ts
```diff
@@ -134,11 +134,15 @@ class WebSocketSSHProxy {
...
         let pipePromise: Promise<void> | undefined;
         session.onAuthenticating((e) => {
+            // TODO vscode_desktop_local_ssh connecting as status
             e.authenticationPromise = this.authenticateClient(e.username ?? '')
                 .then(pipeSession => {
+                    // TODO vscode_desktop_local_ssh connected as status
                     pipePromise = session.pipe(pipeSession);
                     return {};
                 }).catch(async error => {
+                    // TODO vscode_desktop_local_ssh failed as status with coarse grained code
```
akosyakov, line 144: there is also try ... catch before, so maybe more places has to be covered

src/local-ssh/proxy.ts
```diff
@@ -133,14 +129,37 @@ class WebSocketSSHProxy {
...
+                                this.logService.error(e, 'failed to send failed flow status');
+                            });
+                        })(),
+                        this.sendErrorReport(flow, err, 'failed to authenticate proxy with username: ' + e.username)
+                    ]);
+                    this.logService.error(err, 'failed to authenticate client with username: ' + e.username);
+                    await session.close(SshDisconnectReason.byApplication, err.toString(), err instanceof Error ? err : undefined);
                     return null;
                 });
         });
```
akosyakov, line 165: @mustard-mh what is about code below? what is try - catch there?

mustard-mh: > what is try - catch there?

What do you mean? the try catch with `this.authenticateClient`? it's piping the results, return `{}` means accept, `null` means reject if you check document of `e.authenticationPromise`

mustard-mh: > what is about code below?

I have no idea what it (catch) will be after this [hack code](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/2271e81660378696bbd66605390b7a84943f0f03/src/local-ssh/proxy.ts#L114-L122) need to test it

akosyakov: I mean do we need to cover https://github.com/gitpod-io/gitpod-vscode-desktop/blob/2271e81660378696bbd66605390b7a84943f0f03/src/local-ssh/proxy.ts#L166-L176 or it is not related to initial connection?

we for sure should collect or exceptions in GCP since we don't have logging

mustard-mh: I have no idea how to trigger it, `gp stop` will not make produce `this.logService.error(e, 'failed to connect to client'); ` logs.

> or it is not related to initial connection?
It's after connection, to pipe stream

> we for sure should collect or exceptions in GCP since we don't have logging
Yes we can report error here

mustard-mh: > We will remove logging and put proper error reporting instead in another PR or this one?

We're not logging anything, it's a `NopeLogger`, when we want to debug (get logs), we will use `DebugLogger` but it will not in production

mustard-mh: Addressed, feel free to resolve conversation or have further common 🙏

akosyakov: ok, then let's go and check that all error logs places or catches report to GCP in this PR 🙏  (in local proxy)

mustard-mh: > ok, then let's go and check that all error logs places or catches report to GCP in this PR 🙏 (in local proxy)

They are 👍

mustard-mh:
@jeanp413 Could you review this PR and 💄  ?

src/remoteConnector.ts
```diff
@@ -684,6 +684,8 @@ export class RemoteConnector extends Disposable {
 				location: vscode.ProgressLocation.Notification
 			},
 			async () => {
+				// TODO(telemetry): it should be after connecting analytics event
```
akosyakov, line 687: @mustard-mh it is not done yet

src/services/localSSHService.ts
```diff
@@ -60,6 +65,10 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
             this.isSupportLocalSSH = true;
         } catch (e) {
             this.logService.error(e, 'failed to copy local ssh client.js');
+            if (e.message) {
+                e.message = `Failed to copy local ssh client.js: ${e.message}`;
```
akosyakov, line 69: should not we always set it to make easy to identify in error reporting?

src/remote.ts
```diff
@@ -34,13 +34,28 @@ export class NoSSHGatewayError extends Error {
...
         const [, sshDestStr] = remoteUri.authority.split('+');
+        const sshDestInfo = JSON.parse(Buffer.from(sshDestStr, 'hex').toString()) as { hostName: string; user?: string };
         const connectionInfo = context.globalState.get<SSHConnectionParams>(`${SSH_DEST_KEY}${sshDestStr}`);
         if (connectionInfo) {
-            return { remoteAuthority: remoteUri.authority, connectionInfo };
+            let connType: SSHConnType = 'local-app';
+            const domain = getLocalSSHDomain(connectionInfo.gitpodHost);
+            if (sshDestInfo.hostName.endsWith('.' + domain)) {
+                connType = 'local-ssh';
+            } else if (!sshDestInfo.hostName.includes('.')) {
```
jeanp413, line 55: shouldn't we check for `.ssh.` instead? '.' feels trivial

src/remote.ts
```diff
@@ -34,13 +34,28 @@ export class NoSSHGatewayError extends Error {
...
+    connType: SSHConnType;
+}
+
+export function getGitpodRemoteWindowConnectionInfo(context: vscode.ExtensionContext): RemoteConnectionInfo | undefined {
     const remoteUri = vscode.workspace.workspaceFile || vscode.workspace.workspaceFolders?.[0].uri;
     if (vscode.env.remoteName === 'ssh-remote' && context.extension.extensionKind === vscode.ExtensionKind.UI && remoteUri) {
         const [, sshDestStr] = remoteUri.authority.split('+');
+        const sshDestInfo = JSON.parse(Buffer.from(sshDestStr, 'hex').toString()) as { hostName: string; user?: string };
         const connectionInfo = context.globalState.get<SSHConnectionParams>(`${SSH_DEST_KEY}${sshDestStr}`);
         if (connectionInfo) {
```
jeanp413, line 50: doesn't `connectionInfo` already have host?

src/local-ssh/proxy.ts
```diff
@@ -133,14 +129,38 @@ class WebSocketSSHProxy {
...
                     return {};
-                }).catch(async error => {
-                    this.logService.error(error, 'failed to authenticate client with username: ' + e.username);
-                    await session.close(SshDisconnectReason.byApplication, error.toString(), error instanceof Error ? error : undefined);
+                }).catch(async err => {
+                    await Promise.allSettled([
+                        (async () => {
+                            if (err instanceof FailedToProxyError) {
+                                flow.flowFailureCode = err.failureCode;
+                            }
```
jeanp413, line 155: move outside to avoid (()=>{})()

src/remoteConnector.ts
mustard-mh: I'm not clear about it. Right now we fallback to use local-app if `copy file` or `write ssh config` failed, but if we move it  inside, then we will not have fallback just failed? @akosyakov

src/services/localSSHService.ts
mustard-mh: The error been caught will be something like: failed to copy-file / chmod / config ssh, to make it more clear, I add this prefix.

src/remote.ts
mustard-mh: if with local-app, we don't have any '.' but with `.ssh`, it may have a Dedicated host with `ssh`? @jeanp413 @akosyakov

src/remote.ts
mustard-mh: > doesn't connectionInfo already have host?

It doesn't https://github.com/gitpod-io/gitpod-vscode-desktop/blob/064833275fc9e65187a12a8a5cefb442ddc3befd/src/remote.ts#L10-L16

src/remoteConnector.ts
akosyakov: I don't want to change the logic only make sure that connecting is reported before and followed either by failed or success.

akosyakov: We can also another event specific to local ssh configuration flow. `vscode_desktop_local_ssh_config` which is fired once either with success or failure, and then we can have failure/total ratio.

src/remote.ts
jeanp413: local-app has port right? you can get it from `sshDestInfo` then

jeanp413: no idea :thinking: leave it up to you then, but let's rename `local-ssh` to something else is no longer local-ssh maybe `ssh-tunnel` or `ssh-proxy`

src/local-ssh/ipc/extensionServiceServer.ts
```diff
@@ -121,22 +121,19 @@ class ExtensionServiceImpl implements ExtensionServiceImplementation {
...
     // TODO remove from protocol, don't pass sensitive info back and forth, only once for auth, daemon should do telemetry directly
     async sendLocalSSHUserFlowStatus(request: SendLocalSSHUserFlowStatusRequest, _context: CallContext): Promise<{}> {
+        if (!request.flowStatus || request.flowStatus === '') {
+            return {};
+        }
         const flow: UserFlowTelemetry = {
-            flow: 'ssh',
-            kind: 'local-ssh',
-            connType: request.connType === SendLocalSSHUserFlowStatusRequest_ConnType.CONN_TYPE_SSH ? 'ssh' : 'tunnel',
+            flow: 'local_ssh',
```
jeanp413, line 128: same let's rename to `ssh_tunnel` or `ssh_proxy`

src/remoteConnector.ts
mustard-mh: > I don't want to change the logic only make sure that connecting is reported before and followed either by failed or success.

We can't do it, since this logic is way we can identify what we are going to connect with: `local-app` or `local-ssh`, otherwise, we will have two connecting event once fallback

> We can also another event specific to local ssh configuration flow. vscode_desktop_local_ssh_config which is fired once either with success or failure, and then we can have failure/total ratio.

Another event sounds good

src/remote.ts
mustard-mh: Resolved by identify it in `RemoteConnector`

mustard-mh:
Fixed some bugs, and tested with telemetry local loggings. It's good to merge.

But if you want to rename local-ssh, feel free to do it (I prefer to have `local-ssh` actually).

cc @akosyakov @jeanp413

src/local-ssh/ipc/extensionServiceServer.ts
akosyakov: I don't know why we decided to rename 🤷‍♂️  local ssh seems to be more descriptive than other two, we have other ssh proxies like SSH gateway.

akosyakov:
@mustard-mh it seems all TODOs were addressed. I did not test, but I rely on it for you. I cannot approve my own PR, so feel free to continue.

Renaming can be done later.

PR #64, opened 2023-05-16
Pre-release branch, `pre`

filiptronicek:
## Description

Creates a convention where `master` gets synced every night to the `pre` branch, from which a new (nightly) version of the extension is published, inspired by [GitLens](https://github.com/gitkraken/vscode-gitlens). 

## Related Issue(s)

Succeeds https://github.com/gitpod-io/gitpod-vscode-desktop/pull/40
https://gitpod.slack.com/archives/C045J13P8A1/p1684171765423839

## How to test
Run the workflow from https://github.com/gitpod-io/gitpod-vscode-desktop/tree/ft/nightly-prereleases. The `pre` branch should update with any `master` changes and if there none, the workflow should not continue.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -53,7 +53,7 @@ jobs:
   nightly:
     runs-on: ubuntu-latest
     needs: check
-    if: needs.check-release.outputs.status == 'changed' || github.event.inputs.patchVersion
+    if: needs.check.outputs.status == 'changed'
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -5,10 +5,55 @@ on:
...
+          else
+            echo "No pre branch. Creating."
+            git checkout -b pre
+            git push origin pre
+          fi
+          echo "status=changed" >> $GITHUB_OUTPUT
   nightly:
     runs-on: ubuntu-latest
+    needs: check
+    if: needs.check-release.outputs.status == 'changed' || github.event.inputs.patchVersion
```
jeanp413, line 56: Fix typo and shouldn't we always check uf there was any change even if `patchVersion` is provided :thinking: 
```suggestion
    if: needs.check.outputs.status == 'changed'
```

PR #65, opened 2023-05-17
Use createTelemetryLogger api

jeanp413:
## Description

Use createTelemetryLogger api

Related https://linear.app/gitpod/issue/IDE-76/

## How to test

- [x] Check events and error reporting works as expected

You can check telemetry events are logged in the `Extension telemetry` output channel (needs vscode log level set to trace)
image


src/services/telemetryService.ts
```diff
@@ -3,104 +3,22 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { AppenderData, BaseTelemetryClient, BaseTelemetryReporter, RawTelemetryEventProperties, TelemetryEventProperties } from '../common/telemetry';
-import { Analytics, AnalyticsSettings } from '@segment/analytics-node';
 import * as os from 'os';
 import * as vscode from 'vscode';
```
mustard-mh, line 7: We need a telemetry service or common telemetry file without `vscode` imported, then we can use telemetry directly in `proxy.js`

jeanp413: If we decide to send telemetry in proxy itself, then we can just use the library directly, it's straightforward, but will need to send the segment key to it so we will need another request? feels just adding another point of failure is not worth it

mustard-mh: > just inline it?

We can pass an args in ProxyCommand the includes a JSON file stored in extension global

> then we can just use the library directly

It's because we always reuse IXXXServices, but if we don't want to change too much, yes we can use the lib directly

jeanp413: > We can pass an args in ProxyCommand the includes a JSON file stored in extension global

I expect to not expose it like that :grimacing: 

> It's because we always reuse IXXXServices, but if we don't want to change too much, yes we can use the lib directly

yep but proxy.js is the exception as it's independent and not running in the extension host

akosyakov: It can be a bit different code and segment key can be inlined at the build time.

src/services/telemetryService.ts
```diff
@@ -115,34 +29,167 @@ interface TelemetryOptions {
...
+						if (err) {
+							logService.error('Failed to log event to app analytics:', err);
+						}
+					});
+				},
+				sendErrorData: (error, data) => {
+					const properties = data ?? {};
+
+					// Unhandled errors have no data so use host from config
+					const gitpodHost = properties['gitpodHost'] ?? Configuration.getGitpodHost();
```
mustard-mh, line 83: gitpodHost is vscode.TelemetryTrustedValue

mustard-mh (changes requested):

src/services/telemetryService.ts
mustard-mh: What's the benefit to use `vscode.TelemetryTrustedValue`?

src/services/telemetryService.ts
```diff
@@ -115,34 +29,167 @@ interface TelemetryOptions {
...
+					const userId = properties['userId'] ?? '';
+
+					delete properties['gitpodHost'];
+					delete properties['workspaceId'];
+					delete properties['instanceId'];
+					delete properties['userId'];
+
+					const jsonData = {
+						component: 'vscode-desktop-extension',
+						errorStack: error.stack || String(error),
```
mustard-mh, line 101: Error stack is full now, is it expected? I thought we will filter user data out first at before, i.e. user home path/email?

mustard-mh: i.e. deleted `telemetry.ts` file methods `anonymizeFilePaths` and `removePropertiesWithPossibleUserInfo`

src/services/telemetryService.ts
jeanp413: Ah I see it's only unwrapped for logEvent not for error, [there's a TODO](https://github.com/microsoft/vscode/blob/44ef5cc53127cbaa11dee1728bdf8c24522f8fa0/src/vs/workbench/api/common/extHostTelemetry.ts#L278-L279) about it :facepalm:

`vscode.TelemetryTrustedValue` means it won't be anonymized

PR #66, opened 2023-05-19
Workspace management

jeanp413:
## Description

Depends on https://github.com/gitpod-io/gitpod/pull/17682


src/commands/workspaces.ts
```diff
@@ -0,0 +1,172 @@
...
+		// TODO: getWorkspace API need to return path to open, for now harcode it
+		await vscode.commands.executeCommand(
+			'vscode.openFolder',
+			vscode.Uri.parse(`vscode-remote://ssh-remote+${sshDest.toRemoteSSHString()}/workspace/${wsData.repo}`),
+			{ forceNewWindow: false }
+		);
+	}
+}
+
+export class StopEnvironmentCommand implements Command {
```
akosyakov, line 102: Can we don't use Envronment. It is something which we don't have in other code, let's stick to usual terminology, i.e. `StopWorkspace` or `StopWorkspaceInstance`?

PR #68, opened 2023-05-24
Disable error report if extension is not packaged by GHA IDE-108

mustard-mh:
## How to test

- Package extension locally with `yarn package` and you can throw exception somewhere using `telemetryException`
- Install extension locally
- See if there are logs about it with prefix `Local error report`


akosyakov (approved):
code looks sensible, i did not try

PR #69, opened 2023-05-25
Send proper telemetry when copy needed files

mustard-mh:
## Description

> We enabled local ssh config for all users, we should do it only if local ssh experiment is enabled.
> Tracking of analytics event don't match to number of failed exceptions, let's fire event and exception at the same point, i.e. initialized.

- Enable LocalSSHService only when we have feature flag enabled
- Fix update tmp folder partly [[1](https://github.com/gitpod-io/gitpod-vscode-desktop/pull/69/files#diff-d9ccc73eceac9997c215de8b66762f732c45d56bd5ed1fffcd7a70ef66abf176R691-R692)]

## Related Issue(s)

Related [IDE-97](https://linear.app/gitpod/issue/IDE-97)
Fixed [IDE-128](https://linear.app/gitpod/issue/IDE-128), [IDE-100](https://linear.app/gitpod/issue/IDE-100), [IDE-132](https://linear.app/gitpod/issue/IDE-132)

## How to test

- Run extension from source
- It should failed if you change `remote.ssh.configFile` to a unaccessable location

Test result
image

image


mustard-mh:
Code is ready to review, will test later

akosyakov:
It would be nice to land it rather sooner that we don't try to reconfigure all users as now. cc @jeanp413 could you have a look at changes please 🙏

src/services/localSSHService.ts
```diff
@@ -44,12 +47,31 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
...
+            return this.initialized;
+        }
+        this.initialized = this.initialize();
+    }
+
+    private async ensureInitialize() {
+        if (!(await this.experimentService.getUseLocalSSHProxy())) {
+            return;
+        }
+        if (this.initialized) {
```
akosyakov, line 71: I'm not sure it is right way doing it. 2 parallel calls will blcok on previous one and will run in parallel

probably you want a queue, something like that:
```
this.initialized = (async () => {
   await this.initialized;
   this.initialize();
});
```

src/services/localSSHService.ts
```diff
@@ -44,12 +48,31 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
...
                 }
-                if (this.initialized) {
-                    await this.initialized;
-                }
-                this.initialized = this.initialize();
+                await this.initWhenFlagEnabled();
             }
         }));
+        this.experimentService.onDidChangeFlag(async () => {
+            await this.initWhenFlagEnabled();
```
mustard-mh, line 55: It will not work because then setting changed, there is no `userId` so I need to wait until we have userId here, but not sure if it's good. cc @jeanp413 

```
[warning] ConfigCat: Evaluating getValue('gitpod_desktop_use_local_ssh_proxy'). UserObject missing! You should pass a UserObject to getValue(), in order to make targeting work properly. Read more: https://configcat.com/docs/advanced/user-object
```

mustard-mh:
|Open with flag `ON` and then `OFF`| OFF then ON| Change Host to unknown| To Gitpod again|
|:-:|:-:|:-:|:-:|
|image| image |image|image |

mustard-mh:
Test results posted ☝️

src/experiments.ts
```diff
@@ -30,41 +34,48 @@ export class ExperimentalSettings extends Disposable {
...
-            });
+            this.configcatClient = this.createClient(gitpodHost, key, clientLogger);
         }));
-
         this.extensionVersion = new semver.SemVer(extensionVersion);
     }
 
+    createClient(gitpodHost: string, key: string, clientLogger: configcatcommon.IConfigCatLogger) {
+        let ignoredFirstChange = false;
+        return configcat.getClient(key, configcat.PollingMode.AutoPoll, {
```
mustard-mh, line 57: We need to change polling mode to `auto` otherwise hook on `configChanged` makes no sense.

src/services/localSSHService.ts
```diff
@@ -44,12 +48,32 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
...
+
+    public async tryInitialize() {
+        if (this.initialized) {
+            return this.initialized;
+        }
+        this.initialized = this.initialize();
+    }
+
+    private async initWhenFlagEnabled() {
+        await this.sessionService.signIn(this.hostService.gitpodHost);
```
mustard-mh, line 67: Configcat only works when we provide userId [[1](https://github.com/gitpod-io/gitpod-vscode-desktop/pull/69/files#r1205711127)] in most cases. (Only when it enabled/disabled all users not need it)

src/configuration.ts
```diff
@@ -13,13 +13,6 @@ function getGitpodHost() {
     return vscode.workspace.getConfiguration('gitpod').get<string>('host', 'https://gitpod.io/');
 }
 
-function getUseLocalApp(useLocalSSHServer?: boolean) {
-    if (useLocalSSHServer) {
-        return false;
-    }
-    return vscode.workspace.getConfiguration('gitpod').get<boolean>('remote.useLocalApp', false);
```
jeanp413, line 20: Revert this except, so all flags are fetched the same way, wxcept useLocalSSHServer the param we can remove that

src/experiments.ts
jeanp413: why do we need `configChanged` event?

jeanp413:
Don't think this need to be this many changes, just make the `LocalSSHService.initialize` public and call it inside the `sessionService.didFirstLoad` in `extension.ts` (checking the flag with the experiments), cache the promise so when call it in remoteConnector is doesn't copy again

jeanp413:
This feels a little too complicated, how about this:
- Just suscribe to sessionService.didFirstLoad and sessionService.onDidChangeSession and reinitialize there, don't need any changes in experiments.ts

mustard-mh:
I changed to code to @jeanp413 
- Remove feature flag watch
- Send copy file telemetry inside LocalSSHService
- It still do copy every time user start vscode

Going to check if more small changes are needed in Linear

package.json
```diff
@@ -185,7 +185,7 @@
 		"@microsoft/dev-tunnels-ssh-keys": "^3.11.8",
 		"@microsoft/dev-tunnels-ssh-tcp": "^3.11.8",
 		"@segment/analytics-node": "^1.0.0-beta.24",
-		"configcat-node": "^8.0.0",
+		"configcat-node": "^9.1.1",
```
jeanp413, line 188: Let's update in another PR or pushing directly right now it shows me some method is deprecated 
image

src/services/localSSHService.ts
```diff
@@ -12,21 +12,54 @@ import { IHostService } from './hostService';
...
     public isSupportLocalSSH: boolean = false;
     public initialized: Promise<void>;
     constructor(
         private readonly context: vscode.ExtensionContext,
         private readonly hostService: IHostService,
         private readonly telemetryService: ITelemetryService,
-        private readonly logService: ILogService
+        private readonly sessionService: ISessionService,
+        private readonly logService: ILogService,
+        public flow?: UserFlowTelemetry,
```
jeanp413, line 62: remove from constructor and make it a normal property

src/services/localSSHService.ts
```diff
@@ -12,21 +12,54 @@ import { IHostService } from './hostService';
...
 export interface ILocalSSHService {
+    flow?: UserFlowTelemetry;
     isSupportLocalSSH: boolean;
     initialized: Promise<void>;
 }
 
+type FailedToInitializeCode = 'RaceCondition' | 'UnavailableToImprove' | 'UserConfiguredWrong' | 'Unknown';
+
+// TODO: i18n support? 
+const FailedMsgRegexArr: Array<{ regexList: RegExp[]; failureCode: FailedToInitializeCode }> = [
```
akosyakov, line 27: Could we rather have original error codes, like `noSpaceLeft`, `eperm`, `eaccess`, `sshConfigDontExist` and so on. Analytics don't have high cardinality issues and it would be good to preserve original errors.

package.json
mustard-mh: I'll downgrade to 8.0.0

src/remoteConnector.ts
```diff
@@ -687,10 +687,12 @@ export class RemoteConnector extends Disposable {
 				this.usePublicApi = await this.experiments.getUsePublicAPI(params.gitpodHost);
 				this.logService.info(`Going to use ${this.usePublicApi ? 'public' : 'server'} API`);
 
-				let useLocalSSH = await this.experiments.getUseLocalSSHProxy(params.gitpodHost);
+				let useLocalSSH = await this.experiments.getUseLocalSSHProxy();
 				if (useLocalSSH) {
+					// we need to update the remote ssh config first, since another call is too late for local-ssh
+					await this.updateRemoteSSHConfig(true, undefined);
```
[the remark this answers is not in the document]
mustard-mh: We have had check if remote config is updated by local app inside `updateRemoteSSHConfig`, so it's safe

src/services/localSSHService.ts
mustard-mh: Both `fs error` and `vscode.FileSystemError` contain `code` property, going to use it

mustard-mh:
Test result
image

src/services/localSSHService.ts
```diff
@@ -12,21 +12,28 @@ import { IHostService } from './hostService';
...
+import { ISessionService } from './sessionService';
+import { WrapError } from '../common/utils';
 
 export interface ILocalSSHService {
+    flow?: UserFlowTelemetry;
     isSupportLocalSSH: boolean;
     initialized: Promise<void>;
 }
 
+type FailedToInitializeCode = 'RaceCondition' | 'Unknown' | string;
```
akosyakov, line 25: outdated?

src/services/localSSHService.ts
```diff
@@ -85,14 +99,18 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
...
+            const [proxyScript, launcher] = await Promise.all([
+                this.copyFileToGlobalStorage('out/local-ssh/proxy.js', 'sshproxy/proxy.js'),
+                isWindows ? this.copyFileToGlobalStorage('out/local-ssh/proxylauncher.bat', 'sshproxy/proxylauncher.bat') : this.copyFileToGlobalStorage('out/local-ssh/proxylauncher.sh', 'sshproxy/proxylauncher.sh')
+            ]);
+            if (!isWindows) {
+                await fsp.chmod(launcher, 0o755);
+            }
+            return { proxyScript, launcher };
+        } catch (e) {
+            throw new WrapError('Failed to copy local ssh proxy scripts', e, 'RaceCondition');
```
akosyakov, line 112: I am not sure we should put here RaceCondition, why not propagate original error code? What if we are mistaken about race condition?

src/services/localSSHService.ts
```diff
@@ -53,18 +60,25 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
...
+            if (e?.code) {
+                failureCode = e.code;
+            }
+            if (e?.message) {
+                e.message = `Failed to initialize: ${e.message}`;
             }
             this.telemetryService.sendTelemetryException(this.hostService.gitpodHost, e);
             this.isSupportLocalSSH = false;
         }
+        const flowData = this.flow ? this.flow : { gitpodHost: this.hostService.gitpodHost, userId: this.sessionService.safeGetUserId() };
```
akosyakov, line 80: Where this.flow is set? Do we need it?

src/services/localSSHService.ts
```diff
@@ -53,18 +60,25 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
...
+        this.telemetryService.sendUserFlowStatus(this.isSupportLocalSSH ? 'success' : 'failure', { ...flowData, flow: 'local_ssh_config', failureCode });
```
akosyakov, line 81: Could we add info about local app somehow? i.e. whether it is configured, we though that some exceptions are only triggered in case with local app?

src/services/localSSHService.ts
mustard-mh: Then we need a prefix or different event name. Because we can't identify if error comes from copy scripts or configure ssh config.

Most case copy scripts errors are not harmful (except first time to copy)

src/services/localSSHService.ts
```diff
@@ -53,18 +60,25 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
...
-                e.message = `Failed to copy local ssh client.js: ${e.message}`;
+            this.logService.error(e, 'failed to initialize');
+            failureCode = 'Unknown';
+            if (e?.code) {
+                failureCode = e.code;
+            }
+            if (e?.message) {
+                e.message = `Failed to initialize: ${e.message}`;
             }
             this.telemetryService.sendTelemetryException(this.hostService.gitpodHost, e);
```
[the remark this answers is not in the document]
mustard-mh: We can't.
> - When extension active -> we don't have flow here
> - When remote connector handleUri with local-ssh feature flag enabled -> we do have flow here
> 
> source https://github.com/gitpod-io/gitpod-vscode-desktop/pull/69#discussion_r1211417796

If so we need experimentService here, but without UserID, experimentService is useless.

- When extension active -> we can't get UserID so can't see if it's enabled
- When remote connector handleUri with local-ssh feature flag enabled -> it's enabled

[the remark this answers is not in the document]
akosyakov: @mustard-mh I cannot find where it is addressed?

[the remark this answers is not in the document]
mustard-mh: @akosyakov 🙏  https://github.com/gitpod-io/gitpod-vscode-desktop/compare/c86ccda2453554ffe57facd203704cf0b917d5ad...95b15694c2e28cff9806357c070f8e7bf99ef1b8

PR #70, opened 2023-06-02
Bring telemetry inside proxy

mustard-mh:
## Description

- Move telemetry inside proxy

## Related Issue(s)

Fixes [IDE-142](https://linear.app/gitpod/issue/IDE-142)

## Documentation

/hold

### DIFF
--- a/src/local-ssh/telemetryService.ts
+++ b/src/local-ssh/telemetryService.ts
@@ -1,0 +1,331 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Gitpod. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as os from 'os';
+import { Analytics, AnalyticsSettings } from '@segment/analytics-node';
+import { cloneAndChange, escapeRegExpCharacters, mixin } from '../common/utils';
+import { ILogService } from '../services/logService';
+import { getDaemonVersion, getSegmentKey } from './utils';
+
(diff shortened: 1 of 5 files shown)


src/local-ssh/telemetryService.ts
```diff
@@ -0,0 +1,331 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Gitpod. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as os from 'os';
+import { Analytics, AnalyticsSettings } from '@segment/analytics-node';
+import { cloneAndChange, escapeRegExpCharacters, mixin } from '../common/utils';
```
akosyakov, line 8: I wonder there is no way to pull common code from here under `common/telemetry` to reuse between extension and proxy code?

mustard-mh: I do want to do it https://github.com/gitpod-io/gitpod-vscode-desktop/pull/65#discussion_r1197391854 , but we should remove `import 'vscode'` otherwise `proxy.ts` will not work. And it seems the code is not that good to do it.

akosyakov: yeah, I mean pull super class or something, leave template methods where vscode is needed?

mustard-mh: PR https://github.com/gitpod-io/gitpod-vscode-desktop/pull/72 to bring same code into common. But don't know if @jeanp413  like it or not

mustard-mh:
Close to continue with https://github.com/gitpod-io/gitpod-vscode-desktop/pull/72

PR #72, opened 2023-06-05
Bring telemetry inside proxy IDE-142

mustard-mh:
## Description

- Add common telemetry https://github.com/gitpod-io/gitpod-vscode-desktop/pull/70#discussion_r1217735382
- Bring telemetry inside proxy itself

## Related Issue(s)

Fixes #

## How to test

✅  See test results here https://github.com/gitpod-io/gitpod-vscode-desktop/pull/72#issuecomment-1576538418

- Add code below to `extension.ts` and `proxy.ts`, and change logger in proxy.ts
```
telemetryService.sendTelemetryEvent('hello>>>>>>>', { gitpodHost: 'https://gitpod.io', a: '11111'});
telemetryService.sendTelemetryException(new Error('test aa /Users/hwen/aaa/aaa.log'), { gitpodHost: 'https://gitpod.io', a: '22222' });
```
- Run `yarn package` and install `*.visx` file locally
- Reload extension, check its output
- Open Gitpod with VSCode Destkop, check `/tmp/lssh.log` file

## Documentation

/hold


mustard-mh:
Telemetry in extension should work
image

Telemetry in proxy.ts should work
image

src/common/telemetry.ts
```diff
@@ -0,0 +1,293 @@
+/*---------------------------------------------------------------------------------------------
```
mustard-mh, line 1: We don't provide `AbstractTelemetryService` because `telemetryService` in extension need to extend `Disposable`

src/common/vscodeEvent.ts
jeanp413, line 1: let's rename to just `event.ts` as there's no other event class at the moment

webpack.config.js
```diff
@@ -9,6 +9,7 @@ const CopyPlugin = require('copy-webpack-plugin');
 
 const daemonVersion = new webpack.DefinePlugin({
 	'process.env.DAEMON_VERSION': JSON.stringify(packageJSON.daemonVersion ?? '0.0.1'),
+	'process.env.SEGMENT_KEY': JSON.stringify(packageJSON.segmentKey ?? ''),
```
jeanp413, line 12: Do we still need `DAEMON_VERSION`?  can we just inline `packageJSON.version` instead?

src/common/telemetry.ts
```diff
@@ -0,0 +1,293 @@
...
+
+export function createSegmentAnalyticsClient(logService: ILogService, gitpodHost: string, segmentKey: string): Analytics | undefined {
+    const serviceUrl = new URL(gitpodHost);
+
+    const settings: AnalyticsSettings = {
+        writeKey: segmentKey,
+        // in dev mode we report directly to IDE playground source
+        host: 'https://api.segment.io',
+        path: '/v1/batch'
+    };
```
jeanp413, line 53: By default segment buffer the events and then send them when the buffer is full/some time pass, this is fine in the extension because we can flush in the `deactivate` method, for proxy.ts we should disable the buffer and send the events immediately as is the ssh client that manages the lifecycle

mustard-mh:
Pushed a change to make sure we add common properties to event report and omit sensitive data (user path, I thought `password` will be omitted too xd)

image

image

mustard-mh:
Going to merge PR

PR #73, opened 2023-06-06
Ping extension ipc server before use local ssh IDE-77

mustard-mh:
## Related Issue(s)

Relates [IDE-77](https://linear.app/gitpod/issue/IDE-77)

## Documentation

/hold

### DIFF
--- a/src/local-ssh/ipc/extensionServiceServer.ts
+++ b/src/local-ssh/ipc/extensionServiceServer.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ExtensionServiceDefinition, ExtensionServiceImplementation, GetWorkspaceAuthInfoRequest, GetWorkspaceAuthInfoResponse, SendErrorReportRequest, SendLocalSSHUserFlowStatusRequest } from '../../proto/typescript/ipc/v1/ipc';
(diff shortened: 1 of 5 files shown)


mustard-mh:
✅  It can send error report and telemetry if failed to ping
image

✅ It can access workspaces if ping succeed

@jeanp413

mustard-mh:
I'm going to merge this PR

P.S. it adds another telemetry event which will send only when failed `ping_extension_server`

cc @akosyakov @jeanp413

src/local-ssh/ipc/extensionServiceServer.ts
```diff
@@ -198,3 +203,27 @@ export class ExtensionServiceServer extends Disposable {
...
+        server.once('error', () => {
+            resolve(true);
+        });
+        server.once('listening', () => {
+            server.close();
+            resolve(false);
+        });
+        server.listen(port);
+    });
+}
```
jeanp413, line 219: vscode also uses this method but it has this comment ` Uses listen instead of connect. Is faster, but if there is another listener on 0.0.0.0 then this will take 127.0.0.1 from that listener.` do you think we should take it in consideration?
https://github.com/microsoft/vscode/blob/e686621e2ed038f1d21e61f29dd45bc5fdfa680a/src/vs/base/node/ports.ts#L150-L191

mustard-mh: > Uses listen instead of connect. Is faster, but if there is another listener on 0.0.0.0 then this will take 127.0.0.1 from that listener.

It is listen now. Not sure what this mean.

(I tested that listen and ping are very fast 😂)

mustard-mh: We use `127.0.0.1` because extension server listened to it @jeanp413

PR #74, opened 2023-06-09
separate local ssh and gateway (ssh flow)

akosyakov:
## Related Issue(s)

Fixes #

## Documentation

/hold


mustard-mh (approved):
Code LGTM.

- It can connect with local-ssh ✅ 
- It will fallback to gateway and report telemetry when local-ssh port used ✅ 

| Failed as port been used | Fallback to gateway|
|:-:|:-:|
|image |image|
- It will use local-app if user override it ✅ 
image

PR #75, opened 2023-06-11
Improve workspace restart

jeanp413:
## Description

Improve workspace restart

## Related issue
Fixes [IDE-95](https://linear.app/gitpod/issue/IDE-95)


src/workspaceState.ts
```diff
@@ -9,24 +9,61 @@ import * as vscode from 'vscode';
...
         private readonly sessionService: ISessionService,
-        private readonly logger: ILogService,
+        private readonly logService: ILogService,
     ) {
         super();
 
-        this.logger.trace(`WorkspaceState manager for workspace ${workspaceId} started`);
+        this.logService.trace(`WorkspaceState manager for workspace ${workspaceId} started`);
 
         const { onStatusChanged, dispose } = this.sessionService.getAPI().workspaceStatusStreaming(workspaceId);
```
akosyakov, line 68: We can miss phases because of reconnections and general mixups with order, what happens if never see stopping but stopped after reconnection?

src/workspaceState.ts
```diff
@@ -9,24 +9,61 @@ import * as vscode from 'vscode';
...
 
 export class WorkspaceState extends Disposable {
     private workspaceState: WorkspaceStatus | undefined;
+    private _contextUrl: string | undefined;
 
     private _onWorkspaceStateChanged = this._register(new vscode.EventEmitter<void>());
     readonly onWorkspaceStateChanged = this._onWorkspaceStateChanged.event;
 
-    readonly onWorkspaceStopped = filterEvent(this.onWorkspaceStateChanged, () => this.workspaceState?.instance?.status?.phase === WorkspaceInstanceStatus_Phase.STOPPING); // assuming stopping state is never skipped
+    readonly onWorkspaceRunning = onceEvent(filterEvent(this.onWorkspaceStateChanged, () => this.isWorkspaceRunning));
```
akosyakov, line 23: I will rather have one emitter and fire a number which correspond to order of phase, so client can check `ws.phase > RUNNING`. It will cover stopping and stopped.

src/services/remoteService.ts
```diff
@@ -194,4 +199,36 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
...
+            const ws = this.context.globalState.get<WorkspaceRestartInfo>(k)!;
+            this.context.globalState.update(k, undefined);
+            if (new URL(flow.gitpodHost).host === new URL(ws.gitpodHost).host) {
+                this.showWsNotRunningDialog(ws, { ...flow, workspaceId: ws.workspaceId, gitpodHost: ws.gitpodHost });
+            }
+        }
+    }
+
+    private async showWsNotRunningDialog({ workspaceId, gitpodHost, remoteUri }: WorkspaceRestartInfo, flow: UserFlowTelemetryProperties) {
+        const msg = `Workspace ${workspaceId} is not running. Please restart the workspace.`;
```
akosyakov, line 225: I wonder what can we show instead here? Right now it is not very helpful. workspace id does not provide much contextual information. In browser we have an URL to the original context on the stopping screen and info about dirty changes, so a user can instead how is valuable is this workspace or can be rejected.

I wonder whether maybe we can show webview instead with UI similar to browser but styled with Gitpod theme. It should be another PR though.

akosyakov:
@jeanp413 I wonder can you use https://github.com/gitpod-io/gitpod/blob/a18361e5a09b5940a7cd5935802f16eb7f5be5b7/components/supervisor-api/status.proto#L23 instead. It is guarantee to finish gracefully, since supervisor gives 15 second before shutting down the workspace itself. If SSH server is shutdown before we can change it here: https://github.com/gitpod-io/gitpod/blob/a18361e5a09b5940a7cd5935802f16eb7f5be5b7/components/supervisor/pkg/supervisor/supervisor.go#LL469C7-L469C7 but maybe it is already a case

See how browser is using it: https://github.com/gitpod-io/gitpod/blob/a18361e5a09b5940a7cd5935802f16eb7f5be5b7/components/supervisor/frontend/src/ide/supervisor-service-client.ts#L39-L76

akosyakov:
I think it is OK but we need to sort out some issues:
- we should make sure that events are reliably delivered and in proper order
- if some event is missed because of reconnect or so we should take it into account, i.e. don't expect to see stopping
- we should investigate a race with stopping a workspace and ssh server, maybe supervisor hook can be of help here

src/workspaceState.ts
```diff
@@ -35,30 +72,17 @@ export class WorkspaceState extends Disposable {
 
     public async initialize() {
         const ws = await this.sessionService.getAPI().getWorkspace(this.workspaceId);
```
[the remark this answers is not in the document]
jeanp413: order should be taken care of server side, so supervisor was buggy all this time?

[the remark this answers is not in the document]
akosyakov: they are eventually consistent, if order is not important then it is alright

akosyakov:
> it's how vscode executes actions, stream api always report event first, not something supervisor api can help

Why we did not see it with 167 POC?

PR #78, opened 2023-06-13
Add SLOs reports for local ssh IDE-154

mustard-mh:
## Related Issue(s)

Fixes [IDE-154](https://linear.app/gitpod/issue/IDE-154)

## Documentation

/hold

### DIFF
--- a/src/local-ssh/proxy.ts
+++ b/src/local-ssh/proxy.ts
@@ -14,6 +14,7 @@ import * as stream from 'stream';
 import { ILogService } from '../services/logService';
 import { TelemetryService } from './telemetryService';
 import { ITelemetryService, UserFlowTelemetryProperties } from '../common/telemetry';
+import { LocalSSHMetricsReporter } from '../services/localSSHMetrics';
 
 // This public key is safe to be public since we only use it to verify local-ssh connections.
 const HOST_KEY = 'LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1JR0hBZ0VBTUJNR0J5cUdTTTQ5QWdFR0NDcUdTTTQ5QXdFSEJHMHdhd0lCQVFRZ1QwcXg1eEJUVmc4TUVJbUUKZmN4RXRZN1dmQVVsM0JYQURBK2JYREsyaDZlaFJBTkNBQVJlQXo0RDVVZXpqZ0l1SXVOWXpVL3BCWDdlOXoxeApvZUN6UklqcGdCUHozS0dWRzZLYXV5TU5YUm95a21YSS9BNFpWaW9nd2Vjb0FUUjRUQ2FtWm1ScAotLS0tLUVORCBQUklWQVRFIEtFWS0tLS0tCg==';
@@ -77,6 +78,7 @@ class WebSocketSSHProxy {
     constructor(
         private readonly options: ClientOptions,
         private readonly telemetryService: ITelemetryService,
+        private readonly metricsReporter: LocalSSHMetricsReporter,
         private readonly logService: ILogService
     ) {
         this.flow = {
@@ -249,6 +251,7 @@ class WebSocketSSHProxy {
     }
 
     async sendUserStatusFlow(status: 'connected' | 'connecting' | 'failed') {
+        this.metricsReporter.reportConnectionStatus(status, this.flow.failureCode);
         this.telemetryService.sendUserFlowStatus(status, this.flow);
     }
 
@@ -292,7 +295,9 @@ const telemetryService = new TelemetryService(
     logService
 );
 
-const proxy = new WebSocketSSHProxy(options, telemetryService, logService);
+const metricsReporter = new LocalSSHMetricsReporter(options.host);
+
+const proxy = new WebSocketSSHProxy(options, telemetryService, metricsReporter, logService);
 proxy.start().catch(() => {
     // Noop, catch everything in start method pls
 });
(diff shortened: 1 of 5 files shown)


mustard-mh:
Metric report of `local_ssh_config` and `ping` ✅ 
|Succeed|Failure|
|:-:|:-:|
|image|image|


Metric report of `local_ssh` ✅ 
|Succeed|Failure|
|:-:|:-:|
|image|image|

src/services/telemetryService.ts
```diff
@@ -71,12 +71,9 @@ export class TelemetryService extends Disposable implements ITelemetryService {
 	private getSegmentAnalyticsClient(gitpodHost: string, segmentKey: string): Analytics | undefined {
 		const serviceUrl = new URL(gitpodHost);
 		if (this.analitycsClients.has(serviceUrl.host)) {
-			return this.analitycsClients.get(serviceUrl.host)!;
+			return this.analitycsClients.get(serviceUrl.host);
 		}
 		const client = createSegmentAnalyticsClient({ writeKey: segmentKey }, gitpodHost, this.logService);
-		if (!client) {
```
mustard-mh, line 77: Fix nit problem that it will always try to `createSegmentAnalyticsClient` if map is not set. Then for dedicated, it will log error https://github.com/gitpod-io/gitpod-vscode-desktop/blob/18f8a32f3b2d878acec979a03add53015121b1d6/src/common/telemetry.ts#L56

src/services/localSSHMetrics.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    constructor(
+        gitpodHost: string,
+        private readonly logService: ILogService,
+    ) {
+        const serviceUrl = new URL(gitpodHost);
+        this.metricsHost = `ide.${serviceUrl.hostname}`;
+    }
+
+    reportConfigStatus(status: 'success' | 'failure', failureCode?: string) {
```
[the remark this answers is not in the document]
mustard-mh: Addressed. But can be 💄 as it should inside `addCounter`

src/services/localSSHMetrics.ts
```diff
@@ -0,0 +1,61 @@
...
+        this.metricsHost = `ide.${serviceUrl.hostname}`;
+        this._gitpodHost = host;
+    }
+
+    get gitpodHost(): string {
+        return this._gitpodHost;
+    }
+
+    async reportConfigStatus(gitpodHost: string, status: 'success' | 'failure', failureCode?: string): Promise<void> {
+        if (gitpodHost !== this.gitpodHost) {
```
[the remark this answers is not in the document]
akosyakov: or would hostservice always rely on remote window host despite setting already?

src/common/patchFetch.ts
jeanp413, line 1: revert this, as the comment says this is only need in the extension for connect library, in proxy just import `fetch` module

src/services/localSSHMetrics.ts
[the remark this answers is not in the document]
mustard-mh: Addressed 👍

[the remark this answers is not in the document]
mustard-mh: We always get gitpodHost from `hostService`, so it's alright then (will be better until we have multiple host supporting)

src/common/patchFetch.ts
mustard-mh: Then I need two files to make it work in proxy.ts. one to use global fetch (extension.ts), one to import fetch (proxy.ts).

So just keep it? @jeanp413

src/services/localSSHService.ts
```diff
@@ -107,6 +113,7 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
                 await this.configureSettings(locations);
             });
 
+            this.metricsReporter.reportConfigStatus(this.hostService.gitpodHost, 'success');
```
akosyakov, line 116: `this.flow.gitpodHost` ? and the same below

src/services/localSSHService.ts
```diff
@@ -75,6 +79,7 @@ export class LocalSSHService extends Disposable implements ILocalSSHService {
     async extensionServerReady(): Promise<boolean> {
         try {
             await canExtensionServiceServerWork();
+            this.metricsReporter.reportPingExtensionStatus(this.hostService.gitpodHost, 'success');
```
akosyakov, line 82: should not we pull out flow at the begging of the method and then use `flow.gitpodHost` and below as well

src/local-ssh/proxy.ts
```diff
@@ -249,6 +252,9 @@ class WebSocketSSHProxy {
     }
 
     async sendUserStatusFlow(status: 'connected' | 'connecting' | 'failed') {
+        this.metricsReporter.reportConnectionStatus(this.options.host, status, this.flow.failureCode).catch(_e => {
```
akosyakov, line 255: `this.flow.gitpodHost`?

mustard-mh: They are the same

src/common/patchFetch.ts
jeanp413: nope just import `fetch` in common/* files

src/local-ssh/proxy.ts
mustard-mh: Alright, good if we keep use gitpodHost from flow everywhere

akosyakov: true, within the proxy it cannot vary within the same process?

mustard-mh: > within the proxy it cannot vary within the same process?

Yes, `gitpodHost` in proxy.ts is passed by ssh proxy command itself which is constant in one proxy process

PR #79, opened 2023-06-14
Change gitpod config location IDE-165

mustard-mh:
## Related Issue(s)

Relates [IDE-165](https://linear.app/gitpod/issue/IDE-165)

## How to test

✅  See https://github.com/gitpod-io/gitpod-vscode-desktop/pull/79#issuecomment-1590682496

1️⃣  It can replace old content
- Have previsous setup or use previous extension version
- Package and install PR version
- open workspace, and check `~/.ssh/config`(use code_gitpod.d), `~/.ssh/code_gitpod.d/config`(create), `~/.ssh/gitpod/config` (deleted)

2️⃣  It can add new content

3️⃣  It should replace old content and don't delete `~/.ssh/gitpod` folder if it contains another file

4️⃣  It should not delete `~/.ssh/gitpod/config` if it's not generated by us

## Documentation

/hold

### DIFF
--- a/src/ssh/sshConfig.ts
+++ b/src/ssh/sshConfig.ts
@@ -134,17 +134,22 @@ Include "code_gitpod.d/config"
                 throw new WrapError(`Could not read ssh config file at ${configPath}`, e);
             }
         }
+        let hasIncludeTarget = false;
         const oldContent = content;
         const scopeRegex = new RegExp(`## START GITPOD INTEGRATION.+END GITPOD INTEGRATION`, 'sg');
         const matchResult = content.match(scopeRegex);
         if (matchResult) {
-            const matchContent = matchResult[0];
-            if (matchContent !== gitpodIncludeSection) {
-                content = content.replace(scopeRegex, gitpodIncludeSection);
-                // try to check and delete old file
-                tryDeleteOldMatch(matchContent, gitpodHeader);
+            for (const matchContent of matchResult) {
+                if (matchContent !== gitpodIncludeSection) {
+                    content = content.replace(matchContent, '');
+                    // try to check and delete old file
+                    tryDeleteOldMatch(matchContent, gitpodHeader);
+                } else {
+                    hasIncludeTarget = true;
+                }
             }
-        } else {
+        }
+        if (!hasIncludeTarget) {
             content = `${gitpodIncludeSection}\n\n${content}`;
         }
         if (content !== oldContent) {
--- a/src/ssh/sshConfig.ts
+++ b/src/ssh/sshConfig.ts
@@ -120,40 +120,11 @@ export default class SSHConfiguration {
         }
     }
 
-    static async ensureIncludeGitpodSSHConfig(): Promise<void> {
+    private static async addIncludeToUserSSHConfig(gitpodHeader: string): Promise<void> {
         const gitpodIncludeSection = `## START GITPOD INTEGRATION
 ## This section is managed by Gitpod. Any manual changes will be lost.
 Include "code_gitpod.d/config"
 ## END GITPOD INTEGRATION`;
-
-        const gitpodHeader = `### This file is managed by Gitpod. Any manual changes will be lost.`;
-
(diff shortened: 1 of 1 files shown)


src/ssh/sshConfig.ts
```diff
@@ -114,11 +137,20 @@ Include "gitpod/config"
...
-        if (!scopeRegex.test(content)) {
+        const oldContent = content;
+        const scopeRegex = new RegExp(`## START GITPOD INTEGRATION.+END GITPOD INTEGRATION`, 'sg');
+        const matchResult = content.match(scopeRegex);
+        if (matchResult) {
+            const matchContent = matchResult[0];
+            if (matchContent !== gitpodIncludeSection) {
+                content = content.replace(scopeRegex, gitpodIncludeSection);
+                // try to check and delete old file
+                tryDeleteOldMatch(matchContent, gitpodHeader);
```
akosyakov, line 148: Could it be that we will delete by mistake some real user files?

mustard-mh: I have add content check inside this function

(check if it contains header `### This file is managed by Gitpod. Any manual changes will be lost.`

src/ssh/sshConfig.ts
```diff
@@ -114,11 +137,20 @@ Include "gitpod/config"
...
+            }
+        } else {
             content = `${gitpodIncludeSection}\n\n${content}`;
-
+        }
+        if (content !== oldContent) {
```
akosyakov, line 153: Should not we update SSH config only if we managed to copy our files? It does not make sense to include if we fail to copy?

mustard-mh: > Should not we update SSH config only if we managed to copy our files? It does not make sense to include if we fail to copy?

Right now it is.
image

akosyakov: I meant `code_gitpod.d` first, if we cannto then there is no point to include it?

akosyakov: It is a reason why we have broken include now for Chris in first place?

akosyakov:
@filiptronicek do you think you can give it a try on windows machine? Does it handle folder paths with dots nicely?

src/ssh/sshConfig.ts
mustard-mh: I see, you mean to try to write `code_gitpod.d/config` first ? OK, make sense

mustard-mh:
|Test Case|Before|After| It can Connect|
|:-:|:-:|:-:|:-:|
| replace old config |image|image| ✅ |
| all new | image| image | ✅ |
| custom `gitpod/config` file |image |image| ✅ |
| (chris) failed to write gitpod/config as gitpod is a file | image|image| ✅  |

akosyakov (approved):
Code wise and discussions with @mustard-mh it looks sensible, i did not try.

mustard-mh:
> @mustard-mh I know reverted to @jeanp413 POC vsix (old behaviour) and got

@akosyakov  Yes, we need to cover this case. It should work next time you upgrade to new version

akosyakov:
@mustard-mh It would be good enough for me if you test it 🙏

src/ssh/sshConfig.ts
```diff
@@ -100,11 +123,37 @@ export default class SSHConfiguration {
     static async ensureIncludeGitpodSSHConfig(): Promise<void> {
         const gitpodIncludeSection = `## START GITPOD INTEGRATION
 ## This section is managed by Gitpod. Any manual changes will be lost.
-Include "gitpod/config"
+Include "code_gitpod.d/config"
 ## END GITPOD INTEGRATION`;
 
         const gitpodHeader = `### This file is managed by Gitpod. Any manual changes will be lost.`;
 
+        const gitpodConfigFileDir = path.dirname(gitpodSSHConfigPath);
```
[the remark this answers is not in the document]
mustard-mh: Addressed 
image

PR #80, opened 2023-06-15
Proper category errors from serverApi and publicApi IDE-162

mustard-mh:
## Related Issue(s)

Fixes [IDE-162](https://linear.app/gitpod/issue/IDE-162)

## How to test

✅ https://github.com/gitpod-io/gitpod-vscode-desktop/pull/80#issuecomment-1592464464

Add break points and test with cases below

- Server API with random workspaceID
- Server API without auth token
- Public API with random workspaceID
- Public API without auth token

## Documentation

/hold

### DIFF
--- a/src/internalApi.ts
+++ b/src/internalApi.ts
@@ -132,5 +132,5 @@ function categorizeRPCError(err?: ResponseError<any>): Code | undefined {
 	if (err.code >= 400 && err.code < 500) {
 		return Code.InvalidArgument;
 	}
-	return;
+	return Code.Internal;
 }
--- a/src/internalApi.ts
+++ b/src/internalApi.ts
@@ -5,11 +5,13 @@
 
 import { GitpodClient, GitpodServer, GitpodServiceImpl } from '@gitpod/gitpod-protocol/lib/gitpod-service';
 import { JsonRpcProxyFactory } from '@gitpod/gitpod-protocol/lib/messaging/proxy-factory';
-import { listen as doListen } from 'vscode-ws-jsonrpc';
+import { listen as doListen, ResponseError } from 'vscode-ws-jsonrpc';
 import WebSocket, { ErrorEvent } from 'ws';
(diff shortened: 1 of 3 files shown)


mustard-mh:
- Server API with random workspaceID
  image
- Server API without auth token
 image2
- Public API with random workspaceID
  image
- Public API without auth token
  image

src/internalApi.ts
```diff
@@ -84,9 +91,46 @@ class GitpodServerApi extends vscode.Disposable {
...
+			return Code.ResourceExhausted;
+		case 470:
+			return Code.PermissionDenied;
+		case -32603:
+			return Code.Internal;
+	}
+	if (err.code >= 400 && err.code < 500) {
+		return Code.InvalidArgument;
+	}
+	return;
```
akosyakov, line 135: what if err.code >= 500 shold not we return internal?

akosyakov (approved):
let's try

src/internalApi.ts
mustard-mh: https://github.com/gitpod-io/gitpod/blob/d41a38ba83939856e5292e30912f52e749787db1/components/public-api-server/pkg/proxy/errors.go#LL25C1-L26C1

👍 Will do it

src/publicApi.ts
```diff
@@ -66,11 +67,24 @@ export class GitpodPublicApi extends Disposable implements IGitpodAPI {
...
+                }
+                throw err;
+            }
+        };
 
         const transport = createConnectTransport({
             baseUrl: serviceUrl.toString(),
             httpVersion: '2',
-            interceptors: [authInterceptor, metricsInterceptor],
+            interceptors: [errorWrapInterceptor, authInterceptor, metricsInterceptor],
```
mustard-mh, line 87: `errorWrap` should be loaded before `metrics` in case they break each other

PR #88, opened 2023-07-17
Use normalizedContextUrl

mustard-mh:
## Description

- Related https://github.com/gitpod-io/gitpod/pull/18292

## Related Issue(s)

Fixes #

## Documentation

/hold

### DIFF
--- a/src/publicApi.ts
+++ b/src/publicApi.ts
@@ -300,14 +300,18 @@ export interface WorkspaceData {
     phase: WorkspacePhase;
     description: string;
     lastUsed: Date;
-    recentFolders : string[];
+    recentFolders: string[];
 }
 
 export function rawWorkspaceToWorkspaceData(rawWorkspaces: Workspace): WorkspaceData;
 export function rawWorkspaceToWorkspaceData(rawWorkspaces: Workspace[]): WorkspaceData[];
 export function rawWorkspaceToWorkspaceData(rawWorkspaces: Workspace | Workspace[]) {
     const toWorkspaceData = (ws: Workspace) => {
-        const url = new URL(ws.context!.contextUrl);
+        if (ws.context?.details.case !== 'git') {
+            // impossible since we filter them first
+            throw new Error(`Workspace ${ws.workspaceId} is not a git workspace`);
+        }
+        const url = new URL(ws.context.details.value.normalizedContextUrl);
         const provider = url.host.replace(/\..+?$/, ''); // remove '.com', etc
         const matches = url.pathname.match(/[^/]+/g)!; // match /owner/repo
         const owner = matches[0];
@@ -317,7 +321,7 @@ export function rawWorkspaceToWorkspaceData(rawWorkspaces: Workspace | Workspace
             owner,
             repo,
             id: ws.workspaceId,
-            contextUrl: ws.context!.contextUrl,
+            contextUrl: ws.context.details.value.normalizedContextUrl,
             workspaceUrl: ws.status!.instance!.status!.url,
             phase: WorkspaceInstanceStatus_Phase[ws.status!.instance!.status!.phase ?? WorkspaceInstanceStatus_Phase.UNSPECIFIED].toLowerCase() as WorkspacePhase,
             description: ws.description,


src/publicApi.ts
```diff
@@ -300,14 +300,15 @@ export interface WorkspaceData {
...
-    recentFolders : string[];
+    recentFolders: string[];
 }
 
 export function rawWorkspaceToWorkspaceData(rawWorkspaces: Workspace): WorkspaceData;
 export function rawWorkspaceToWorkspaceData(rawWorkspaces: Workspace[]): WorkspaceData[];
 export function rawWorkspaceToWorkspaceData(rawWorkspaces: Workspace | Workspace[]) {
     const toWorkspaceData = (ws: Workspace) => {
-        const url = new URL(ws.context!.contextUrl);
+        const contextUrl = ws.context?.details.case === 'git' ? ws.context.details.value.normalizedContextUrl : ws.context.contextUrl;
```
akosyakov: I don't think we should ever use context URL on UI, but either:
- ensure that Public API always return normalized context URL according to the logic in https://github.com/gitpod-io/gitpod/blob/0bb6aaad9d4923d5510de865437f22c7f726833f/components/gitpod-protocol/src/context-url.ts#L31C28-L31C28
- or call to https://github.com/gitpod-io/gitpod/blob/0bb6aaad9d4923d5510de865437f22c7f726833f/components/gitpod-protocol/src/context-url.ts#L31C28-L31C28 from here 

we should also filter out or return undefined if it cannot be recovered

akosyakov: i.e ws.context.contextUrl should be used as is

mustard-mh: > ensure that Public API always return normalized context URL according to the logic in

Public API defined different context, like `prebuild`, `git`, `snapshot`. But only use `git` [[1](https://github.com/gitpod-io/gitpod/blob/7e0c605a3d470b8cee0e841e51da9b20022f4f4b/components/public-api-server/pkg/apiv1/workspace.go#L332-L335)] right now.

We can filter that it only return `git` type and use `normalizedContextUrl`? To call server protocol again looks weird

mustard-mh: `context URL` is a part of UI now, if we remove this option, then we can remove those properties

image

akosyakov: For now to work around maybe the easiest, call https://github.com/gitpod-io/gitpod/blob/0bb6aaad9d4923d5510de865437f22c7f726833f/components/gitpod-protocol/src/context-url.ts#L31C28-L31C28 pass fake objet without normalizedContextUrl but context url. You have gitpod-protocol already in the extension.

Later we need to talk with @geropl @svenefftinge to align how we want to deal context URLs. It maybe we should rather parse in server metadata from context URL, store them and fetch here instead of trying to derive from the URL at all.

src/publicApi.ts
```diff
@@ -320,7 +318,7 @@ export function rawWorkspaceToWorkspaceData(rawWorkspaces: Workspace | Workspace
             owner,
             repo,
             id: ws.workspaceId,
-            contextUrl,
+            contextUrl: url.toString(),
```
akosyakov, line 321: it can be null as far as I understand, we will fail in such case here, I think we should adjust clients to be safe

akosyakov: i.e. just filter such

mustard-mh:
It's hard to say it works with backward compatible (`ContextURL.getNormalizedURL` does return ok), but at least UI won't break
image

image

cc @akosyakov

src/publicApi.ts
```diff
@@ -300,14 +301,31 @@ export interface WorkspaceData {
...
+            } else {
+                const normalized = ContextURL.getNormalizedURL({ context: {}, contextURL: ws.context!.contextUrl } as any);
+                if (!normalized) {
+                    return undefined;
+                }
+                url = normalized;
+            }
+        } catch (e) {
+            // ignore
+            return undefined;
```
akosyakov, line 327: it will be unexpected? probably better to log?

akosyakov (approved):

src/publicApi.ts
mustard-mh: I'd like to add, but it seems needs HostService\LogService\TelemetryService here and change lots of files so that we can send exception. Maybe we will need inject in repo? I'd like to keep it and add TODO as it's right now.

cc @jeanp413 for `Maybe we will need inject in repo?` this question

image

jeanp413:
@mustard-mh Why this got merged xD, workaround from Anton should be enough maybe some 💄 but we need to discuss with Sven or Gero first about fix in server side, personally I don't think raw workspace url should be exposed in public api

jeanp413:
I'll revert for now and we can fix properly after public api is updated

akosyakov:
> workaround from Anton should be enough maybe some
@jeanp413 my work ardoun does not fix anything for UI, onyl for one particular path, UI is failing right now whenever you start a workspace from prebuild and so on. Discussions about Public API will take a while, many people are having time off now. And they may not lead to what you expect.

jeanp413:
> @jeanp413 my work ardoun does not fix anything for UI, onyl for one particular path, UI is failing right now whenever you start a workspace from prebuild and so on.

yeah I see, I'm working on change that does not touch public api and instead normalizes on client side by removing prefix

akosyakov:
>yeah I see, I'm working on change that does not touch public api and instead normalizes on client side by removing prefix

We already have this code in gitpod-protocol. Coudl we keep it in one place than implement a custom logic?

jeanp413:
>  I'm kind of confused what was wrong with Huiwen's PR

it's using the `normalizedContextUrl`, and unless it's already decided that we'll keep both `contextUrl` and `normalizedContextUrl` I prefer not to add it for now beucase then we have to mantain it

> We already have this code in gitpod-protocol. Coudl we keep it in one place than implement a custom logic?

I took a look at it but it's buggy as it does not take into account `open-prebuild` prefix, in any case it can be simplified to just search for `http[s]?` as all context parsers rely on `/` as separator [[1](https://github.com/gitpod-io/gitpod/blob/c747c1e9330529fd1caf8502959fa7988e721ac4/components/server/src/prebuilds/start-prebuild-context-parser.ts#L14)] [[2](https://github.com/gitpod-io/gitpod/blob/c747c1e9330529fd1caf8502959fa7988e721ac4/components/server/src/workspace/referrer-prefix-context-parser.ts#L13-L14)] [[3](https://github.com/gitpod-io/gitpod/blob/c747c1e9330529fd1caf8502959fa7988e721ac4/components/server/src/workspace/open-prebuild-prefix-context-parser.ts#L17-L18)] [[4](https://github.com/gitpod-io/gitpod/blob/c747c1e9330529fd1caf8502959fa7988e721ac4/components/server/src/workspace/imagebuild-prefix-context-parser.ts#L13-L14)] [[5](https://github.com/gitpod-io/gitpod/blob/c747c1e9330529fd1caf8502959fa7988e721ac4/components/server/src/workspace/envvar-prefix-context-parser.ts#L36-L37)]

PR #89, opened 2023-07-22
Support ssh gateway when connecting from workspace explorer

jeanp413:
## Description

Support ssh gateway when connecting from workspace explorer

## How to test

1. Disable local ssh in configcat
2. Connect using workspace explorer, it should work

### DIFF
--- a/src/commands/workspaces.ts
+++ b/src/commands/workspaces.ts
@@ -17,6 +17,7 @@ import { ITelemetryService } from '../common/telemetry';
 import { IRemoteService } from '../services/remoteService';
 import { WrapError } from '../common/utils';
 import { getOpenSSHVersion } from '../ssh/sshVersion';
+import { IExperimentsService } from '../experiments';
 
 function getCommandName(command: string) {
 	return command.replace('gitpod.workspaces.', '').replace(/(?:_inline|_context)(?:@\d)?$/, '');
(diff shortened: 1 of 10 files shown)


src/commands/workspaces.ts
```diff
@@ -123,18 +120,42 @@ export class ConnectInNewWindowCommand implements Command {
...
 
-						await this.initializeLocalSSH(wsData!.id);
-
 						await raceCancellationError(eventToPromise(wsState.onWorkspaceRunning), cancelToken);
+						wsData = wsState.workspaceData; // Update wsData with latest info after workspace is running
+					}
+
+					let sshDest: SSHDestination;
+					let password: string | undefined;
+					if (await this.experimentsService.getUseLocalSSHProxy()) {
```
akosyakov, line 129: Should not be here a check that local ssh is initialized? i.e the logic should be similar to open URI.

akosyakov: I see it is initialised above. I find it is quite confusing with so many paths. I think there should be only one function which hides the entires connectivity logic, i.e. we should not call initializeLocalSSH here and there, but only this function can. All other clients start the workspace and then call to this function. It does not need to be the part of this PR, but would be nice to clean up.

akosyakov (approved):
I tested both with Local SSH and SSH Gateway from the browser and view. And everything seems to work.

PR #90, opened 2023-07-24
[local ssh] improve observability and try to prevent hanging

akosyakov:
## Description

- add started/exited status reporting to identify hanging process
- don't swallow errors, but report all, we can sort them out later to investigate root causes of hanging
- add detection of stale web sockets to prevent hanging similar how it is done by Local App

## Related Issue(s)

Fixes #

## Documentation

/hold

### DIFF
--- a/src/local-ssh/proxy.ts
+++ b/src/local-ssh/proxy.ts
@@ -50,23 +50,25 @@ const flow: SSHUserFlowTelemetry = {
     processId: process.pid,
 };
 
-telemetryService.sendUserFlowStatus('started', flow)
-const sendExited = (exitCode: number, forceExit: boolean, exitSignal?: NodeJS.Signals) => telemetryService.sendUserFlowStatus('exited', {
-    ...flow,
-    exitCode,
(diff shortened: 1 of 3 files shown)


src/local-ssh/proxy.ts
```diff
@@ -227,10 +268,45 @@ class WebSocketSSHProxy {
...
+                let pingTimeout: NodeJS.Timeout | undefined;
+                const heartbeat = () => {
+                    stopHearbeat();
+
+                    // Use `WebSocket#terminate()`, which immediately destroys the connection,
+                    // instead of `WebSocket#close()`, which waits for the close timer.
+                    // Delay should be equal to the interval at which your server
+                    // sends out pings plus a conservative assumption of the latency.
+                    pingTimeout = setTimeout(() => {
+                        socket.terminate();
```
akosyakov, line 289: we could instead for now report analytics to see how often it happens

akosyakov (commented):
@jeanp413 to double test, 💄 and land today 🙏

PR #91, opened 2023-07-26
Add local ssh test connection

jeanp413:
## How to test

- [ ] Check there are no regressions
- [ ] Check test connection works, specially on windows


src/ssh/nativeSSH.ts
```diff
@@ -0,0 +1,111 @@
...
+                version = match[0];
+                return version;
+            }
+        }
+    } catch {
+    }
+    return undefined;
+}
+
+export async function getHostConfig(hostname: string) {
```
akosyakov, line 76: is it used anywhere?

src/remoteSession.ts
```diff
@@ -67,7 +67,7 @@ export class RemoteSession extends Disposable {
 		try {
 			const useLocalSSH = await this.experiments.getUseLocalSSHProxy();
```
mustard-mh, line 68: If it failed to test local ssh and with feature flag enabled, then other code may not run

src/ssh/nativeSSH.ts
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Test ssh connection to a remote machine using native ssh client.
+ *
+ * Caller should handle any exception thrown
+ */
+export async function testSSHConnection(username: string, hostname: string) {
+    const sshPath = getSSHConfigPath();
+    const randomId = crypto.randomBytes(12).toString('hex');
+    // TODO: support password input so we can replace the ssh2 library
+    const resp = execCommand(sshPath, ['-T', '-o', 'ConnectTimeout=5', `${username}@${hostname}`, `echo "${randomId}"`], { timeout: 5500 });
```
mustard-mh, line 98: Will `timeout 5s` too short? since we have `10s` of retry in proxy itself, in my network condition, it may cost 3+ seconds (depend on provider state)

src/commands/workspaces.ts
```diff
@@ -124,12 +124,25 @@ export class ConnectInNewWindowCommand implements Command {
...
+						);
+					}
+
 					let sshDest: SSHDestination;
 					let password: string | undefined;
 					if (await this.experimentsService.getUseLocalSSHProxy()) {
-						const domain = getLocalSSHDomain(this.hostService.gitpodHost);
-						const sshHostname = `${wsData!.id}.${domain}`;
-						sshDest = new SSHDestination(sshHostname, wsData!.id);
+						sshDest = localSSHDestination;
```
mustard-mh, line 145: If it failed to test, then we should not use it too

mustard-mh (commented):
Tested it can start with local ssh in Windows.

Left some comments ☝️ 

| Start with LocalSSH| LocalSSH test failed |
|:-:|:-:|
|SCR-20230726-oygt|SCR-20230726-ozbc|

src/ssh/nativeSSH.ts
akosyakov: We are afraid too block unnecessary users on SSH Gateway.

The plan is to rollback windows users to SSH Gateway and then do dry runs. I would suggest to go with small timeout and if we see a lot of timeout issues then we increase it.

PR #93, opened 2023-07-28
Fix channel hanging

jeanp413:
## Description

Added

```
		channel.onEof(() => {
			void PipeExtensions.forwardData(channel, toChannel, Buffer.alloc(0)).catch();
		});
		toChannel.onEof(() => {
			void PipeExtensions.forwardData(channel, toChannel, Buffer.alloc(0)).catch();
		});
```

Though now I see `Fri Jul 28 2023 11:30:21 TRACE: [websocket][9024] Invalid channel ID 10 in {messageString}. undefined` in logs, will investigate more if it's relevant or not


mustard-mh:
With this patch, I can't reproduce this bug now 👍 . 

But did not test others cases, maybe a chance that we need to introduce integration test now

src/local-ssh/proxy.ts
```diff
@@ -26,12 +26,12 @@ if (!options) {
...
 
-import { NopeLogger } from './logger';
-const logService = new NopeLogger();
+// import { NopeLogger } from './logger';
+// const logService = new NopeLogger();
 
 // DO NOT PUSH CHANGES BELOW TO PRODUCTION
-// import { DebugLogger } from './logger';
-// const logService = new DebugLogger();
+import { DebugLogger } from './logger';
```
mustard-mh, line 33: Don't forget to revert this change

mustard-mh:
When it connect, I can find trace loggings like below sometimes, maybe worth to have a look too
```
Sat Jul 29 2023 01:20:43 GMT+0800 (China Standard Time)TRACE: [websocket][9024] Invalid channel ID 6 in f (recipientChannel=6). undefined
```

akosyakov:
Tests are great idea! Can it be done on level of local ssh?

src/local-ssh/pipeSession.ts
```diff
@@ -0,0 +1,192 @@
+import { CancellationToken, PromiseCompletionSource, SessionRequestFailureMessage, SessionRequestMessage, SessionRequestSuccessMessage, SshChannel, SshChannelClosedEventArgs, SshChannelError, SshChannelOpeningEventArgs, SshMessage, SshRequestEventArgs, SshSession, SshSessionClosedEventArgs, SshTraceEventIds, TraceLevel } from "@microsoft/dev-tunnels-ssh";
```
akosyakov, line 1: is it a temporary copy from the upstream? It would nice to have a comment? Or rather fork and use our version?

mustard-mh: Jean Pierre's upstream PR https://github.com/microsoft/dev-tunnels-ssh/pull/65 is approved, it's easier to patch this way. We can delete this file after upstream release

PR #94, opened 2023-07-29
Upgrade connect lib

jeanp413:
https://github.com/bufbuild/connect-es/pull/712 got merged and also added unit tests, let's update and see if we can remove the workaround after this is merged and deployed


mustard-mh:
It seems to back to my origin commit, not sure if it's working, let's try

PR #95, opened 2023-08-08
Use `normalizedContextUrl`

filiptronicek:
## Description

If we can, use the normalized context URL we get now.


jeanp413:
It was discussed `normalizedContextUrl` will be removed so it shouldn't be used at all, instead we should use clone url  and git status, [ref](https://linear.app/gitpod/issue/EXP-89#comment-86dd21bf)

filiptronicek:
Thanks @jeanp413 🙏

PR #97, opened 2023-09-29
Remove local app logic

jeanp413:


src/extension.ts
```diff
@@ -127,8 +127,6 @@ export async function activate(context: vscode.ExtensionContext) {
 		// Because auth provider implementation is in the same extension, we need to wait for it to activate first
 		sessionService.didFirstLoad.then(async () => {
 			if (remoteConnectionInfo) {
-				commandManager.register({ id: 'gitpod.api.autoTunnel', execute: () => remoteConnector.autoTunnelCommand });
```
akosyakov, line 130: let's keep as a dummy thing here but clean up in https://github.com/gitpod-io/gitpod-code/blob/1de61b5711089287b7f27bc1223c57d6d9bb3144/gitpod-remote/src/extension.ts#L49

src/remoteConnector.ts
```diff
@@ -760,12 +345,12 @@ export class RemoteConnector extends Disposable {
 							this.logService.error('No SSH gateway:', e);
 							const ok = 'OK';
 							await this.notificationService.showWarningMessage(`${e.host} does not support [direct SSH access](https://github.com/gitpod-io/gitpod/blob/main/install/installer/docs/workspace-ssh-access.md), connecting via the deprecated SSH tunnel over WebSocket.`, { flow: gatewayFlow, id: 'no_ssh_gateway' }, ok);
```
akosyakov, line 347: The message is wrong now. If both Local SSH and SSH Gateway we should say something like `failed to connect to the workspace` and point out how troubleshooting that users can report bugs or inspect logs

src/remoteConnector.ts
```diff
@@ -760,12 +345,12 @@ export class RemoteConnector extends Disposable {
...
-							// Do nothing and continue execution
+							return undefined;
 						} else if (e instanceof SSHError && e.message === 'Timed out while waiting for handshake') {
 							this.logService.error('SSH test connection error:', e);
 							const ok = 'OK';
 							await this.notificationService.showWarningMessage(`Timed out while waiting for the SSH handshake. It's possible, that SSH connections on port 22 are blocked, or your network is too slow. Connecting via the deprecated SSH tunnel over WebSocket instead.`, { flow: gatewayFlow, id: 'ssh_timeout' }, ok);
```
akosyakov, line 352: we should remove here as well `Connecting via the deprecated SSH tunnel over WebSocket instead` and point to troubleshooting docs too

akosyakov (approved):

src/remoteConnector.ts
akosyakov: We should not point to https://github.com/gitpod-io/gitpod/blob/main/install/installer/docs/workspace-ssh-access.md anymore since we don't have self hosted.

But rather point to Local SSH since it works over HTTP, we should not use local SSH is in user faced messages or proxy or so

src/remoteConnector.ts
akosyakov: we either just skip it or use https://github.com/gitpod-io/gitpod-vscode-desktop/pull/97/files#r1346338466

src/remoteConnector.ts
```diff
@@ -760,12 +345,12 @@ export class RemoteConnector extends Disposable {
...
-							// Do nothing and continue execution
```
akosyakov, line 768: the link below is broken we shoudl point onw to https://www.gitpod.io/docs/references/ides-and-editors/vscode#connecting-to-vs-code-desktop

akosyakov (commented):
we should also delete https://www.gitpod.io/docs/references/ides-and-editors/legacy-vscode-connections#connect-to-vs-code-desktop-using-local-companion and all references to it

src/remoteConnector.ts
```diff
@@ -765,57 +349,15 @@ export class RemoteConnector extends Disposable {
 									if (action === seeLogs) {
 										this.logService.show();
 									} else if (action === showTroubleshooting) {
-										vscode.env.openExternal(vscode.Uri.parse('https://www.gitpod.io/docs/ides-and-editors/vscode#connecting-to-vs-code-desktop-ssh'));
+										vscode.env.openExternal(vscode.Uri.parse('https://www.gitpod.io/docs/references/ides-and-editors/vscode#connecting-to-vs-code-desktop'));
```
akosyakov, line 352: @jeanp413 we should reference `https://www.gitpod.io/docs/references/ides-and-editors/vscode#troubleshooting` here, SSH Gateway is till relevant we cannot omit it here

akosyakov: also left a note here: https://github.com/gitpod-io/website/pull/4001#pullrequestreview-1661373331

jeanp413:
```suggestion
										vscode.env.openExternal(vscode.Uri.parse('https://www.gitpod.io/docs/references/ides-and-editors/vscode#troubleshooting'));
```

PR #98, opened 2023-09-30
Wider vscode support

jeanp413:


src/remoteConnector.ts
```diff
@@ -580,31 +580,12 @@ export class RemoteConnector extends Disposable {
...
-			}
-		} else {
-			// TODO(ak) notify a user about config file changes?
-			if (currentConfigFile === localAppSSHConfigPath) {
-				// invalidate cached SSH targets from the current config file
-				await remoteSSHconfig.update('configFile', undefined, vscode.ConfigurationTarget.Global);
-			}
-			await remoteSSHconfig.update('configFile', localAppSSHConfigPath, vscode.ConfigurationTarget.Global);
+	private async ensureRemoteSSHExtInstalled(flow: UserFlowTelemetryProperties): Promise<boolean> {
+		const isOfficialVscode = vscode.env.uriScheme === 'vscode' || vscode.env.uriScheme === 'vscode-insiders';
```
akosyakov, line 584: Should not we ensure that yours extension is installed then? Should we also collect your extension version then too as part of error reporting/tracing/analytics?

src/services/remoteService.ts
```diff
@@ -177,11 +178,12 @@ export class RemoteService extends Disposable implements IRemoteService {
         await SSHConfiguration.saveGitpodSSHConfig(gitpodConfig);
     }
 
-    private getHostSSHConfig(host: string, launcher: string, proxyScript: string, extIpcPort: number, logLevel:string) {
+    private getHostSSHConfig(host: string, launcher: string, proxyScript: string, extIpcPort: number, logLevel: string) {
+        const extraArgs = (process.versions['electron'] && process.versions['microsoft-build']) ? '--ms-enable-electron-run-as-node' : '';
```
akosyakov, line 182: should not we rather rely on vscode.env.uriScheme === 'vscode' || vscode.env.uriScheme === 'vscode-insiders'?  and encapsualte it somewhere?

I'm not sure about stability of `microsoft-build` version

src/services/remoteService.ts
```diff
@@ -326,6 +328,30 @@ export class RemoteService extends Disposable implements IRemoteService {
         throw new Error('SSH password modal dialog, Canceled');
     }
 
+    async updateRemoteSSHConfig(usingSSHGateway: boolean, localAppSSHConfigPath: string | undefined) {
```
akosyakov, line 331: is it only move, without any semantic changes?

src/remoteConnector.ts
jeanp413: it's tricky because cursor forked my extension so it has another id 🤦

src/services/remoteService.ts
jeanp413: this check is done in vscode source code too so it should be stable I guess

src/services/remoteService.ts
jeanp413: yep

src/remoteConnector.ts
akosyakov: I mean we can have a case if it is not cursor then?

But the best would be that Cursor folks contribute back to your extension and don't cause discrepancy cc @mbrevoort  if you know whom to cc :)

src/services/remoteService.ts
akosyakov: What is drawback not relying on it but on stable uriScheme?

akosyakov (approved):

src/services/remoteService.ts
jeanp413: I think both are reliable, though uriScheme can be changed in product.json manually

PR #99, opened 2023-10-04
Add vscode version and remote ssh version to telemetry

jeanp413:


src/local-ssh/proxy.ts
```diff
@@ -394,13 +380,61 @@ class WebSocketSSHProxy {
...
 
-const metricsReporter = new LocalSSHMetricsReporter(logService);
+let vscodeProductJson: any;
+async function getVSCodeProductJson(appRoot: string) {
+    if (!vscodeProductJson) {
+        try {
+            const productJsonStr = await fs.promises.readFile(path.join(appRoot, 'product.json'), 'utf8');
+            vscodeProductJson = JSON.parse(productJsonStr);
+        } catch {
+            return {};
```
akosyakov, line 390: 👍 good that it does not fail the process

src/local-ssh/proxy.ts
```diff
@@ -34,46 +46,6 @@ if (!options) {
...
-);
-
-const flow: SSHUserFlowTelemetry = {
-    flow: 'local_ssh',
-    gitpodHost: options.gitpodHost,
-    workspaceId: '',
-    processId: process.pid,
-};
-
-telemetryService.sendUserFlowStatus('started', flow);
```
akosyakov, line 57: it is bad that it is not first event anymore, since we cannot see how much time it takes from started to connected, i.e. we cannot tell effect of reading versions on connection performance. Maybe we omit versions then for started?

akosyakov (approved):

PR #100, opened 2023-10-09
Support flexible auth user

jeanp413:


akosyakov (approved):
looks sensible , i did not try

src/proto/ipc/v1/ipc.proto
```diff
@@ -4,76 +4,12 @@ package ipc.v1;
 
 service ExtensionService {
     rpc GetWorkspaceAuthInfo (GetWorkspaceAuthInfoRequest) returns (GetWorkspaceAuthInfoResponse) {}
-    rpc SendLocalSSHUserFlowStatus (SendLocalSSHUserFlowStatusRequest) returns (SendLocalSSHUserFlowStatusResponse) {}
```
akosyakov, line 7: felling uneasy about it, since upgrade happen between any 2 version of extensions, but unimplemented result should not break a flow though?

PR #101, opened 2023-12-07
use http create key endpoint

iQQBot:
Switch to use http endpoint for create private key pair

This endpoint introduced in https://github.com/gitpod-io/gitpod/pull/18411 which is 4 month ago, gitpod cloud and dedicated user already have this endpoint

To use this endpoint, we can be more smoothly compatible with next-gen.

### DIFF
--- a/src/local-ssh/ipc/extensionServiceServer.ts
+++ b/src/local-ssh/ipc/extensionServiceServer.ts
@@ -13,45 +13,16 @@ import { Server, createChannel, createClient, createServer } from 'nice-grpc';
 import { ITelemetryService } from '../../common/telemetry';
 import { Configuration } from '../../configuration';
 import { timeout } from '../../common/async';
-import { BrowserHeaders } from 'browser-headers';
-import { ControlServiceClient, ServiceError } from '@gitpod/supervisor-api-grpcweb/lib/control_pb_service';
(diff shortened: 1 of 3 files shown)


src/local-ssh/ipc/extensionServiceServer.ts
```diff
@@ -13,47 +13,18 @@ import { Server, createChannel, createClient, createServer } from 'nice-grpc';
...
-import { ConnectError, Code } from '@connectrpc/connect';
+import { ConnectError } from '@connectrpc/connect';
 import { rawWorkspaceToWorkspaceData } from '../../publicApi';
 
 function isServiceError(obj: any): obj is ServiceError {
     // eslint-disable-next-line eqeqeq
     return obj != null && typeof obj === 'object' && typeof obj.metadata != null && typeof obj.code === 'number' && typeof obj.message === 'string';
 }
 
-function wrapSupervisorAPIError<T>(callback: () => Promise<T>, opts?: { maxRetries?: number; signal?: AbortSignal }): Promise<T> {
```
akosyakov, line 31: We should still keep retry in case of network error.

meysholdt:
Closing because the PR has been inactive for more than 90 days. Feel free to reopen if needed

PR #102, opened 2024-01-31
Fix local-ssh not work in vscode insider

mustard-mh:
## Description

electron seems remove support for `--ms-enable-electron-run-as-node` flags in `vscode insider` It will cause electron not to swallow this flags, resulting in a change in our parameter position.

image

## Related Issue(s)

Fixes #

## How to test

1. build a extension using this branch and install it in both `vscode insider` and `vscode stable`, and restart vscode
2. try to open workspace via `vscode insider` and `vscode stable`
3. check remote hostname, it should be `*.vsi.*` or `*.vss.*`

## Documentation

/hold

### DIFF
--- a/src/local-ssh/proxy.ts
+++ b/src/local-ssh/proxy.ts
@@ -20,7 +20,7 @@ interface ClientOptions {
 }
 
 function getClientOptions(): ClientOptions {
-    const args = process.argv.slice(2);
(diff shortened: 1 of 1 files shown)


jeanp413:
Ref https://github.com/microsoft/vscode/pull/202053

PR #105, opened 2024-07-04
Collect lssh logs in connection window

jeanp413:
## Description

Collect lssh logs in connection window

/hold

### DIFF
--- a/src/commands/logs.ts
+++ b/src/commands/logs.ts
@@ -23,6 +23,7 @@ interface IFile {
 
 export class ExportLogsCommand implements Command {
 	readonly id = 'gitpod.exportLogs';
+	static latestLSSHHost: string | undefined;
 
 	constructor(
 		private readonly context: vscode.ExtensionContext,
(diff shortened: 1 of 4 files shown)


src/commands/logs.ts
```diff
@@ -23,6 +23,7 @@ interface IFile {
 
 export class ExportLogsCommand implements Command {
 	readonly id = 'gitpod.exportLogs';
+	static latestLSSHHost: string | undefined;
```
mustard-mh, line 26: @jeanp413  Will it make sense to persist last connection state in other place than memory? I tried locally:
- Is connected to workspace -> **has lssh logs**
- Window show disconnected and attempt to retry -> **not able to collect logs**
- Reload window -> **no last time's lssh logs**

I'm not sure where `in connection window` means other than the second point above. But it's not working

jeanp413: it only applies when it's a normal window (not remote) and the user is trying to connect to a workspace and it fails before redirecting to the remote window

mustard-mh: > Window show disconnected and attempt to retry -> not able to collect logs

I thought this case is used to happen to users side as well, but as this PR has its own scope, let's publish it

PR #106, opened 2024-07-10
Improve heartbeat activity detection on `onDidChangeWindowState` event

mustard-mh:
## Description

Official description of `active` property

> Whether the window has been interacted with recently. This will change immediately on activity, or after a short time of user inactivity.

## Related Issue(s)

Fixes [[internal chat](https://gitpod.slack.com/archives/C0201TXMV54/p1720603695491779?thread_ts=1720593853.632269&cid=C0201TXMV54)]

## How to test

No need?

## Documentation

/hold

### DIFF
--- a/src/heartbeat.ts
+++ b/src/heartbeat.ts
@@ -76,8 +76,13 @@ export class HeartbeatManager extends Disposable {
         this._register(vscode.window.onDidCloseTerminal(() => this.updateLastActivity('onDidCloseTerminal')));
         this._register(vscode.window.onDidChangeTerminalState(() => this.updateLastActivity('onDidChangeTerminalState')));
         this._register(vscode.window.onDidChangeWindowState((e) => {
-            this.focused = e.focused;
-            this.updateLastActivity('onDidChangeWindowState');
+            // active property was introduced in 1.89, but not exist in 1.75
+            // https://code.visualstudio.com/api/references/vscode-api#WindowState
+            const isActive: Boolean | undefined = (e as any).active;
+            if (isActive === true || e.focused !== this.focused) {
+                this.focused = e.focused;
+                this.updateLastActivity('onDidChangeWindowState');
+            }
         }));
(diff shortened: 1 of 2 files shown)


src/heartbeat.ts
```diff
@@ -76,8 +76,13 @@ export class HeartbeatManager extends Disposable {
         this._register(vscode.window.onDidCloseTerminal(() => this.updateLastActivity('onDidCloseTerminal')));
         this._register(vscode.window.onDidChangeTerminalState(() => this.updateLastActivity('onDidChangeTerminalState')));
         this._register(vscode.window.onDidChangeWindowState((e) => {
-            this.focused = e.focused;
-            this.updateLastActivity('onDidChangeWindowState');
+            // active property was introduced in 1.89, but not exist in 1.75
```
jeanp413, line 79: You need to increase `"engines": {"vscode": "^1.82.0"},` in package.json and also the vscode types in devDependencies

src/heartbeat.ts
```diff
@@ -76,8 +76,13 @@ export class HeartbeatManager extends Disposable {
...
+            // https://code.visualstudio.com/api/references/vscode-api#WindowState
+            const isActive: Boolean | undefined = (e as any).active;
+            if (isActive === true || e.focused !== this.focused) {
+                this.focused = e.focused;
+                this.updateLastActivity('onDidChangeWindowState');
+            }
```
jeanp413, line 85: With this new api we don't need the other events so let remove them, and send heartbeat only when `isActive` is `true` every `HEARTBEAT_INTERVAL + random value`

src/heartbeat.ts
mustard-mh: It requires `@types/vscode` version `1.89`, then our extension will need to update engines to `^1.89.0`

mustard-mh: Resolve as we don't need to check `active` anymore

jeanp413 (approved):

PR #107, opened 2024-07-11
Avoid stale disconnects when network is slow

jeanp413:
## Description

Avoid stale disconnects when network is slow

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/19284

/hold


mustard-mh (approved):
Tested it can connect to a workspace with lssh

PR #108, opened 2024-07-17
Load custom certificates

jeanp413:
## Description
Load custom certicates from the OS, currently only works on mac and linux, for windows it requires a native binary but I think we can make it work by loading the one from vscode installation

## Related Issue(s)

Fixes https://linear.app/gitpod/issue/ENT-483/


mustard-mh:
Verified custom cert worked nicely on macOS
image

mustard-mh (approved):

PR #113, opened 2024-10-21
Dev Container heartbeats

filiptronicek:
## Description

If a user happens to use the "Reopen in container" feature of VS Code Desktop, we should still be able to send heartbeats and not let workspaces time out.

## How to test

1. Build the extension using `yarn package`
2. Install it 
3. Open a workspace from https://github.com/filiptronicek/vscode-remote-try-go/tree/ft/ak/magic in `catfood`
4. Use "Reopen in Container" when VS Code Desktop asks you
5. Check the Gitpod output to see we're sending heartbeats every now and then

### DIFF
--- a/src/services/remoteService.ts
+++ b/src/services/remoteService.ts
@@ -347,14 +347,11 @@ export class RemoteService extends Disposable implements IRemoteService {
             vscode.workspace.getConfiguration('dev.containers');
         const defaultDevContainerExtConfigInfo = remoteDevContainerConfig.inspect<string[]>('defaultExtensions');
         const defaultDevContainerExtensions = defaultDevContainerExtConfigInfo?.globalValue ?? [];
-        const defaultDCExtConfigInfo =
-            remoteDevContainerConfig.inspect<string[]>('defaultExtensions');
-        const defaultDcExtensions = defaultDCExtConfigInfo?.globalValue ?? [];
         if (!defaultDevContainerExtensions.includes('gitpod.gitpod-remote-ssh')) {
             defaultDevContainerExtensions.unshift('gitpod.gitpod-remote-ssh');
             await remoteDevContainerConfig.update(
                 'defaultExtensions',
-                defaultDcExtensions,
+                defaultDevContainerExtensions,
                 vscode.ConfigurationTarget.Global,
             );
         }
@@ -407,7 +404,7 @@ export class RemoteService extends Disposable implements IRemoteService {
         } catch (error) {
             const msg = `Error while installing local extensions on remote.`;
             this.logService.error(msg);
-            this.logService.trace(error)
+            this.logService.trace(error);
 
             const status = 'failed';
             const seeLogs = 'See Logs';
--- a/src/services/remoteService.ts
+++ b/src/services/remoteService.ts
@@ -331,25 +331,27 @@ export class RemoteService extends Disposable implements IRemoteService {
 
     async updateRemoteConfig() {
         const remoteSSHconfig = vscode.workspace.getConfiguration('remote.SSH');
-        const defaultExtConfigInfo =
+        const defaultSSHExtConfigInfo =
             remoteSSHconfig.inspect<string[]>('defaultExtensions');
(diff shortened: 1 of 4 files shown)


jeanp413:
It's not required to install local extension on the remote correct? or do we need to update `dev.containers.defaultExtensions`?

src/services/remoteService.ts
```diff
@@ -329,19 +329,42 @@ export class RemoteService extends Disposable implements IRemoteService {
...
+            );
+        }
+
+        const remoteDevContainerConfig =
+            vscode.workspace.getConfiguration('dev.containers');
+        const defaultDCExtConfigInfo =
+            remoteDevContainerConfig.inspect<string[]>('defaultExtensions');
+        const defaultDcExtensions = defaultDCExtConfigInfo?.globalValue ?? [];
+        if (!defaultExtensions.includes('gitpod.gitpod-remote-ssh')) {
+            defaultExtensions.unshift('gitpod.gitpod-remote-ssh');
```
jeanp413, line 352: Should be `defaultDcExtensions`

src/services/remoteService.ts
```diff
@@ -329,19 +329,42 @@ export class RemoteService extends Disposable implements IRemoteService {
...
+                vscode.ConfigurationTarget.Global,
+            );
+        }
+
+        const remoteDevContainerConfig =
+            vscode.workspace.getConfiguration('dev.containers');
+        const defaultDCExtConfigInfo =
+            remoteDevContainerConfig.inspect<string[]>('defaultExtensions');
+        const defaultDcExtensions = defaultDCExtConfigInfo?.globalValue ?? [];
+        if (!defaultExtensions.includes('gitpod.gitpod-remote-ssh')) {
```
jeanp413, line 351: Should be `defaultDcExtensions`

jeanp413 (commented):
We also need to install the dev container extension if it's not present, same we for the remote-ssh extension

src/services/remoteService.ts
filiptronicek: Nice catch! Updated in [`ff25f22` (#113)](https://github.com/gitpod-io/gitpod-vscode-desktop/pull/113/commits/ff25f22f51b3e534c4e21c874e13badf16e4e43e) 🙏

akosyakov:
>We also need to install the dev container extension if it's not present, same we for the remote-ssh extension

I'm not sure it is necessary. We don't want to have it generally on Classic, i.e. gitpod.yml still is preferred way. Maybe documentation is enough.

src/services/remoteService.ts
```diff
@@ -329,19 +329,44 @@ export class RemoteService extends Disposable implements IRemoteService {
...
+            );
+        }
+
+        const remoteDevContainerConfig =
+            vscode.workspace.getConfiguration('dev.containers');
+        const defaultDevContainerExtConfigInfo = remoteDevContainerConfig.inspect<string[]>('defaultExtensions');
+        const defaultDevContainerExtensions = defaultDevContainerExtConfigInfo?.globalValue ?? [];
+        const defaultDCExtConfigInfo =
+            remoteDevContainerConfig.inspect<string[]>('defaultExtensions');
+        const defaultDcExtensions = defaultDCExtConfigInfo?.globalValue ?? [];
```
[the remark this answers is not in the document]
filiptronicek: @jeanp413 sorry for the hiccup, must've misread when addressing your other comments. Looks like it's working now:
image

jeanp413 (approved):
LGTM now but turns out `dev.containers.defaultExtensions` only works if the image is build using the devcontainers extension 🤦 , If we use a prebuild image using the devcontainer cli in a gitpod task, it doesn't install the remote extension

filiptronicek:
@jeanp413 yikes! Does that mean that the extension is then included in the image itself? I'm merging now, and let's indeed think about a fix later

jeanp413:
yes in the devcontainer extension the can merge additional config and then it's stored as part of image metadata

PR #120, opened 2025-04-25
Don't use binary format for transport in public api client

jeanp413:
## Description

Don't use binary format, binary format forces http2, it's preferable to downgrade to http1.1 if http2 is not available for some reason 

## Related Issue(s)

Fixes #

## Documentation

/hold


mustard-mh (approved):
image

Smoke tested it can connect using new build

